PR #4, opened 2019-06-26
Support re-entrant nodes

Geod24:


Geod24:
@AndrejMitrovic : It works

source/geod24/LocalRest.d
```diff
@@ -48,6 +48,10 @@ private struct Command
 {
     /// Tid of the sender thread (cannot be JSON serialized)
     C.Tid sender;
+    /// In order to support re-entrancy, every request contains an id
+    /// which should be copied in the `Response`
+    /// Initialized to `size_t.max` so not setting it crashes the program
+    size_t id = size_t.max;
```
AndrejMitrovic, line 54: Where is the code that crashes it if it's equal to size_t.max?

Geod24: The array indexing here: https://github.com/Geod24/localrest/pull/4/files#diff-b5229c55d46fb0a852c49674585a3595R311

source/geod24/LocalRest.d
```diff
@@ -72,6 +81,101 @@ private struct ArgWrapper (T...)
...
+        }
+
+        private bool notified;
+    }
+}
+
+
+/// We need a scheduler to simulate an event loop and to be re-entrant
+/// However, the one in `std.concurrency` is process-global (`__gshared`)
+private LocalScheduler scheduler;
```
AndrejMitrovic, line 176: It's a bit ugly to have it in module scope, but I guess it's complicated to move it inside RemoteAPI because we would have multiple instances of it for every type RemoteAPI is instantiated with..

Geod24: It is ugly but I don't see any way around it.

Geod24:
Submitted some fixes upstream: https://github.com/dlang/phobos/pull/7096

Geod24:
Tested and known to work.

PR #5, opened 2019-06-27
Trigger DMD/LDC/GDC bug

Geod24:
Based on #4, just here for the CI.

### DIFF
--- a/source/geod24/LocalRest.d
+++ b/source/geod24/LocalRest.d
@@ -841,8 +841,6 @@ unittest
     static import std.concurrency;
     import std.format;
 
-    __gshared C.Tid[string] tbn;
-
     static interface API
     {
         @safe:
@@ -862,7 +860,9 @@ unittest
 
         public override void setNext (string name) @trusted
         {
-            this.next = new RemoteAPI!API(tbn[name]);
+            auto tid = std.concurrency.locate(name);
+            assert(tid != typeof(tid).init, name);
+            this.next = new RemoteAPI!API(tid);
         }
 
         private API next;
@@ -875,7 +875,7 @@ unittest
     ];
 
     foreach (idx, ref api; nodes)
-        tbn[format("node%d", idx)] = api.tid();
+        std.concurrency.register(format("node%d", idx), api.tid());
     nodes[0].setNext("node1");
     nodes[1].setNext("node2");
     nodes[2].setNext("node0");


Geod24:
`dub test --build=debug` works...

AndrejMitrovic:
Can this be closed now? The test is incorporated in the other PR right?

Geod24:
No that bug is still here.

ghost:
Problem seems to be in std.concurrency. After the operation is run, it clears the information for the thread. Which removes the registered name. Why it works in debug builds might just be that it is running slower, as it is node[2] that relies on node[0]'s thread not being cleaned up (removing the registered name).

https://github.com/dlang/phobos/blob/v2.087.1/std/concurrency.d#L1438

```D
    void create(void delegate() op) nothrow
    {
        void wrap()
        {
            scope (exit)
            {
                thisInfo.cleanup(); // <--- FiberScheduler's ThreadInfo
            }
            op();
        }

        m_fibers ~= new InfoFiber(&wrap);
    }

```


```D
    RemoteAPI!(API)[3] nodes = [
        RemoteAPI!API.spawn!Node(),
        RemoteAPI!API.spawn!Node(),
        RemoteAPI!API.spawn!Node(),
    ];

    foreach (idx, ref api; nodes)
        std.concurrency.register(format("node%d", idx), api.tid());
    nodes[0].setNext("node1"); // <--- node[0]'s ThreadInfo gets cleared (unregistering all names in std.concurrency)
                               //      when the fiber completes execution
    nodes[1].setNext("node2");
    nodes[2].setNext("node0"); // <--- attempts to use node[0]'s name registered in std.concurrency which was already removed
                               //      if the fiber used above has completed execution
```

Geod24:
Yep.
While debugging this I declared the AA as `pragma(mangle, "...") extern [definition]` so I could easily dump it from the module. IIRC the content of the name was removed on access, which does not make sense.

ghost:
It gets removed in a separate thread. ThreadInfo.cleanup() calls unregisterMe() for the FiberScheduler's thread after one of it's Fibers finishes, which removes it from the AA in std.concurrency. It'd depend on how you debugged it and if the Fiber's thread (initiated by nodes[0].setNext) was allowed to finish executing or not.

Geod24:
Uh, are you saying `FiberScheduler` messes with the registration ?
Yeah, we use a scheme where we have multiple schedulers. But it seems `std.concurrency`, while exposing types that look like they allow it, rely on global state all over the place.

ghost:
FiberScheduler removes all names associated with a ThreadInfo id when any Fiber it created completes. This is because it calls ThreadInfo.cleanup() in wrap(), which is what InfoFiber ends up executing:

https://github.com/dlang/phobos/blob/v2.087.1/std/concurrency.d#L1438
```D
    // in FiberScheduler
    void create(void delegate() op) nothrow
    {
        void wrap()
        {
            scope (exit)
            {
                thisInfo.cleanup(); // removes name from global AA when fiber finishes executing
            }
            op();
        }

        m_fibers ~= new InfoFiber(&wrap);
    }
```

https://github.com/dlang/phobos/blob/v2.087.1/std/concurrency.d#L1117
```D
   // in struct ThreadInfo

    void cleanup()
    {
        if (ident.mbox !is null)
            ident.mbox.close();
        foreach (tid; links.keys)
            _send(MsgType.linkDead, tid, ident);
        if (owner != Tid.init)
            _send(MsgType.linkDead, owner, ident);
        unregisterMe(); // <---------- Removes self from global AA Here
    }
```

https://github.com/dlang/phobos/blob/v2.087.1/std/concurrency.d#L992

```D
private void unregisterMe()
{
    auto me = thisInfo.ident;
    if (thisInfo.ident != Tid.init)
    {
        synchronized (registryLock)
        {
            if (auto allNames = me in namesByTid)
            {
                // ------------- Gets removed from globals here -----------------
                foreach (name; *allNames)
                    tidByName.remove(name);
                namesByTid.remove(me);
            }
        }
    }
}
```

At least this is what is happening with the test code.



Unrelated found another potential bug waiting to happen (it really is a mess of global state)... ThreadInfo.cleanup() calls unregisterMe() which grabs the local threads ThreadInfo to cleanup. But you can call ThreadInfo.cleanup() from a thread that isn't its own. ThreadInfo.cleanup() should probably be static, or cleanup the actual ThreadInfo that it was called with.

Geod24:
> it really is a mess of global state

Yep, and that's probably what happened with Vibe.d in the linked issue on the bounty.
We had to put [a workaround for Vibe.d's own workaround](https://github.com/bpfkorea/agora/blob/3de8d05a6678226e816a0329fb2b7636d491e383/source/agora/test/Base.d#L413-L416).
Also, I already fixed one such bug [in the past](https://github.com/dlang/phobos/pull/7096) and there are probably more looming.

I think the underlying issue is that you can't put fibers and Thread behind the same abstraction (Tasks) unless you deal exclusively with value types, because concurrency is not parallelism.

Nevertheless, I really wish this module was properly working, because it is immensely useful.

CyberShadow:
No longer reproducible as of DMD 2.090.0.

Geod24:
Fixed in https://github.com/dlang/phobos/pull/7314
And we forked `std.concurrency` anyways.

PR #6, opened 2019-06-30
Actually use GDC-9 for testing

Geod24:
Because 'd: dmd' is applied after 'DC=gdc' is exported,
it overrode DC and we ended up testing with DMD.

### DIFF
--- a/source/geod24/LocalRest.d
+++ b/source/geod24/LocalRest.d
@@ -156,8 +156,8 @@ public final class RemoteAPI (API) : API
             SWITCH:
                 switch (cmd.method)
                 {
-                    static foreach (member; __traits(allMembers, API))
-                        static foreach (ovrld; __traits(getOverloads, API, member))
+                    foreach (member; __traits(allMembers, API))
+                        foreach (ovrld; __traits(getOverloads, API, member))
                         {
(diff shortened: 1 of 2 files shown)


source/geod24/LocalRest.d
```diff
@@ -447,3 +432,96 @@ unittest
...
     assert(nodes[0].requests() == 7);
 }
+
+
+/*******************************************************************************
+
+    Applies a template predicate to a list of arguments and merge the generated
+    overload set.
+
+    This template is explicitly design to generate function definitions within
```
AndrejMitrovic, line 442:
```suggestion
    This template is explicitly designed to generate function definitions within
```

source/geod24/LocalRest.d
```diff
@@ -447,3 +432,96 @@ unittest
...
+    a scope and add them to an overload set.
+    The use of `sym` is necessary because overload sets from mixed-in templates
+    are not merged in the parent.
+
+    Params:
+        Pred = Predicate to generate one of more functions named `sym`
```
AndrejMitrovic, line 448:
```suggestion
        Pred = Predicate to generate one or more functions named `sym`
```

I think?

PR #9, opened 2019-07-10
Add the ability to arbitrarily suspend a node

Geod24:
Let's see how slow Travis is!

### DIFF
--- a/source/geod24/LocalRest.d
+++ b/source/geod24/LocalRest.d
@@ -401,17 +401,34 @@ public final class RemoteAPI (API) : API
 
     /***************************************************************************
 
-        Returns the `Tid` this `RemoteAPI` wraps
+        Introduce a namespace to avoid name clashes
 
-        This can be useful for calling `std.concurrency.register` or similar.
-        Note that the `Tid` should not be used directly, as our event loop,
-        would error out on an unknown message.
+        The only way we have a name conflict is if someone exposes `ctrl`,
+        in which case they will be served an error along the following line:
+        LocalRest.d(...): Error: function `RemoteAPI!(...).ctrl` conflicts
+        with mixin RemoteAPI!(...).ControlInterface!() at LocalRest.d(...)
 
     ***************************************************************************/
 
-    public C.Tid tid () @nogc pure nothrow
+    public mixin ControlInterface!() ctrl;
+
+    /// Ditto
+    private mixin template ControlInterface ()
     {
-        return this.childTid;
+        /***********************************************************************
+
+            Returns the `Tid` this `RemoteAPI` wraps
+
+            This can be useful for calling `std.concurrency.register` or similar.
+            Note that the `Tid` should not be used directly, as our event loop,
+            would error out on an unknown message.
+
+        ***********************************************************************/
+
+        public C.Tid tid () @nogc pure nothrow
+        {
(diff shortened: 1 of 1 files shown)


Geod24:
For reference:
```
# Linux + DMD
Two non-blocking calls: 118 μs and 1 hnsec
Sleep + non-blocking call: 35 μs and 7 hnsecs
Delta since sleep: 1 sec, 181 μs, and 4 hnsecs
```

```
# OSX + DMD
Two non-blocking calls: 325 μs and 3 hnsecs
Sleep + non-blocking call: 204 μs and 4 hnsecs
Delta since sleep: 1 sec, 1 ms, 406 μs, and 2 hnsecs
```

```
# Linux + LDC
Two non-blocking calls: 1 ms and 220 μs
Sleep + non-blocking call: 43 μs
Delta since sleep: 1 sec, 224 μs, and 2 hnsecs
All unit tests have been run successfully.
```

```
# OSX + LDC
Two non-blocking calls: 284 μs
Sleep + non-blocking call: 93 μs and 9 hnsecs
Delta since sleep: 1 sec and 956 μs
```

```
# Linux + GDC
Two non-blocking calls: 169 μs and 8 hnsecs
Sleep + non-blocking call: 37 μs and 9 hnsecs
Delta since sleep: 1 sec, 215 μs, and 6 hnsecs
```

Geod24:
That wouldn't be hard to introduce either

Geod24:
If you can state your requirements I can implement it fairly quickly.

Some things I can think of:
- Ask a node to drop the next X messages
- Ask a node to have a failure ratio of Y (e.g. 0.6 => 60% of messages are dropped, randomly)
- Ask a node to ignore messages matching a certain pattern

In addition to those actions ("message filters"), we can ask a node to apply those filters so only a node (or a set of), and during a certain period (expressed in time or in messages).

But that is a lot of flexibility to implement, and the way we handle this might be very dependent on the scenarios we have to tests. My initial idea was that very advanced scenarios (e.g. a Byzantine node) would be handled by a subclass of `Node` (a different `Implementation`) rather than done at `LocalRest` level.

So let's start with what we need.

Geod24:
@AndrejMitrovic : Added the ability to drop message.
I think the next step is to add timeout to all network operations (from the main thread and from nodes).
At the moment "timeout" in nodes mean that some memory will never be free, but functionally it "just works".
However, a message from the main thread being dropped means that there we'll sleep forever. Be aware.

Geod24:
Let me know if that's enough for your use case.

AndrejMitrovic:
This is good for starters. In the future I'd like to be able to just drop N messages, so we don't suffer from random timing failures (also sleeping increases the total time to run the tests).

LGTM.

source/geod24/LocalRest.d
```diff
@@ -365,6 +380,16 @@ public final class RemoteAPI (API) : API
                 throw e;
     }
 
+    /// Clear up `max` messages from the pending messages
+    private static size_t removeMessages (size_t count)
```
AndrejMitrovic, line 384: Oh btw, I assume you're going to use the return value of this in a future PR?

Geod24: Yeah it's something I'm considering. I will add the ability to drop X messages soon ™️

PR #10, opened 2019-07-15
Implement request filtering

AndrejMitrovic:


AndrejMitrovic:
I mean maybe I just have to wrap my failing calls in the localrest unittest in a try/catch..

AndrejMitrovic:
Ah yes. I've fixed the test.

Anyway what really needs good thought is the API itself. Should it be a filter delegate? Or an object? And what parameters should it take, should it just match on the function name or also try to match on the arguments passed?

It could go any different way really in terms of design, there's many possibilities. I tried to keep it simple by just allowing it to match on the function name for now.

The unittests should pass now.

AndrejMitrovic:
So I came up with a different API, it's in the last commit. 

I ran into a small issue with it, take a look:

```D
node.filterAPI!(node.foo);     // won't work, wrong mangling
node.filterAPI!(node.API.foo); // this will work!
```

The first one will actually match the override implemented in `RemoteAPI`, and not the method that's in the base class.

So I've added a `static assert` inside of `filterAPI`.

I think with a little bit of trickery I could make both of the calls work, by stripping out the "RemoteAPI" part in the mangled string. I'm not sure if we want this magic though..

What do you think?

AndrejMitrovic:
Also the naming could be better, "filterAPI" sounds lame.

AndrejMitrovic:
I've improved it further. Now it works automagically: https://github.com/Geod24/localrest/compare/86eb78d142a247e7fcaeae8e06f0fb9e095bec22..b0189ffb00c3eaa5fb0b71938585f605ef8e76a3

Edit: Since there is `ControlInterface!() ctrl;` now, it's easy to disambiguate any name clashes, so I've renamed it to just `filter`.

AndrejMitrovic:
Anyway feel free to throw the code away if you want to implement it better. I'll enable testing it now.

source/geod24/LocalRest.d
```diff
@@ -475,6 +509,51 @@ public final class RemoteAPI (API) : API
         {
             C.send(this.childTid, TimeCommand(d, dropMessages));
         }
+
+        /***********************************************************************
+
+            Filter any requests issued to the provided method.
```
Geod24, line 519:
```suggestion
            Filter any requests issued to the provided method.
            
```

source/geod24/LocalRest.d
```diff
@@ -475,6 +509,51 @@ public final class RemoteAPI (API) : API
...
+            Calling the API endpoint will throw an exception,
+            therefore the request will fail.
+
+            Use via:
+
+            -----
+            interface API { void call(); }
+            class C : API { void call() { } }
+            auto obj = new RemoteAPI!API(...);
+            obj.filter!(obj.call);
```
Geod24: Why not just pass `API.method` here ?

Geod24: Also could you add an example with overloads ?

source/geod24/LocalRest.d
```diff
@@ -475,6 +509,51 @@ public final class RemoteAPI (API) : API
...
+            -----
+
+            The remote node will call `Thread.sleep`, becoming completely
```
Geod24: I don't see such a call ?

source/geod24/LocalRest.d
```diff
@@ -475,6 +509,51 @@ public final class RemoteAPI (API) : API
...
+              method = the method to filter requests from
+
+        ***********************************************************************/
+
+        public void filter (alias method) () @trusted
+        {
+            // ensure it's used with API.method, *not* RemoteAPI.method which
+            // is an override of API.method. Otherwise mangling won't match!
+            enum method_name = __traits(identifier, method);
+            enum mangled = mixin("API." ~ method_name).mangleof;
```
Geod24: I got a feeling this does not support overloads

source/geod24/LocalRest.d
```diff
@@ -896,3 +975,65 @@ unittest
...
+    __gshared C.Tid node_tid;
+
+    static interface API
+    {
+        size_t getCount();
+        void foo ();
+        void asyncFoo ();
+    }
+
+    import std.stdio;
```
Geod24: Leftover?

source/geod24/LocalRest.d
```diff
@@ -896,3 +975,65 @@ unittest
...
+                ()
+                {
+                    try
+                    {
+                        this.remote.foo();
+                    }
+                    catch (Exception ex)
+                    {
+                        import std.algorithm;
+                        assert(ex.msg.startsWith("Filtered command"));
```
Geod24: You could just do an `==` comparison here, no ?

source/geod24/LocalRest.d
```diff
@@ -896,3 +975,65 @@ unittest
...
+        }
+    }
+
+    auto node = RemoteAPI!API.spawn!Node();
+    node_tid = node.tid();
+    node.foo();
+    assert(node.getCount() == 1);
+
+    import std.stdio;
+    bool fooFilter (string cmd) { writefln("Command was: %s", cmd); return cmd == "foo"; }
```
Geod24: Leftover?

Geod24 (approved):
Looks pretty good

source/geod24/LocalRest.d
AndrejMitrovic: > Why not just pass API.method here ?

That works too.

> Also could you add an example with overloads ?

Doesn't support it yet. It would be hard to support this on the call side without making it awkward though. Do you have any ideas?

Normally in D to match a specific overload you'd have to do something like `void function(int, float) foo = &object.foo`, but that's ugly to use.

source/geod24/LocalRest.d
AndrejMitrovic: woops copypasta

source/geod24/LocalRest.d
AndrejMitrovic: It's a bit brittle then. Any kind of tiny change will break this unittest. For example, if you add another field to `Command`. And the `Tid` which is part of the message changes between runs too.

AndrejMitrovic:
Pushed suggested changes. Not sure how to handle overloads best yet.

We could perhaps do:

```
node.filter!(API.call, int, float)
```

Would that look ok to you?

source/geod24/LocalRest.d
Geod24: I meant to compare the messages only. The user code never sees `Command` anyway.

AndrejMitrovic:
Pushed fixes. It's https://github.com/Geod24/localrest/compare/35c8cf71f5d168e317ecb79869569f9db9c9f079..67ca2f4fcba109b4c223ca2f25d42e3c23a0bc1f and https://github.com/Geod24/localrest/compare/67ca2f4fcba109b4c223ca2f25d42e3c23a0bc1f..a6216c4f624ed39d2536d3f69c4b0f37de87192f.

source/geod24/LocalRest.d
```diff
@@ -366,11 +385,26 @@ public final class RemoteAPI (API) : API
...
                         (Response res) {
                             scheduler.pending = res;
                             scheduler.waiting[res.id].c.notify();
                         },
-                        (Command cmd) { scheduler.spawn(() => handleCommand(cmd, node)); });
+                        (Command cmd)
+                        {
+                            if (cmd.method == filt_func)
+                            {
+                                // we have to send back a message
```
Geod24: In fact if you had the `drop` option, you might not have to

source/geod24/LocalRest.d
```diff
@@ -896,3 +1006,106 @@ unittest
...
+
+    // foo() filtered
+    node.asyncFoo();
+    assert(node.fooCount() == 1);  // it was not called!
+
+    node.clearFilter();  // clear the filter
+    node.asyncFoo();
+    assert(node.fooCount() == 2);  // it was called!
+
+    // veriy foo(int) works first
```
```suggestion
    // verify foo(int) works first
```

source/geod24/LocalRest.d
```diff
@@ -896,3 +1006,106 @@ unittest
...
+            );
+        }
+    }
+
+    auto node = RemoteAPI!API.spawn!Node();
+    node_tid = node.tid();
+    node.foo();
+    assert(node.fooCount() == 1);
+
+    bool fooFilter (string cmd) { return cmd == "foo"; }
```
Geod24: ?

source/geod24/LocalRest.d
```diff
@@ -896,3 +1006,106 @@ unittest
...
+
+    // both of these work
+    static assert(is(typeof( node.filter!(API.foo) )));
```
```suggestion
    static assert(is(typeof(node.filter!(API.foo))));
```

source/geod24/LocalRest.d
```diff
@@ -896,3 +1006,106 @@ unittest
...
+    static assert(is(typeof( node.filter!(node.foo) )));
```
```suggestion
    static assert(is(typeof(node.filter!(node.foo))));
```

source/geod24/LocalRest.d
```diff
@@ -896,3 +1006,106 @@ unittest
...
+
+    // veriy foo(int) works first
+    node.asyncFoo(1);
+    assert(node.fooCount() == 2);
+    assert(node.fooIntCount() == 1);  // first time called
+
+    // now filter only the int overload
+    node.filter!(API.foo, int);
+
+    // make sure the () overload is still not filtered
```
```suggestion
    // make sure the argumentless overload is still not filtered
```

Geod24 (approved):
Just a few nits but LGTM

AndrejMitrovic:
Fails on some mac systems. Argh.

AndrejMitrovic:
I'm trying to figure out the cause by commenting out code piecemeal.

But is there any way we can get stack traces to show on a segfault?

AndrejMitrovic:
Ok, I've reduced the problem to the last `asyncFoo()` call in the code.

AndrejMitrovic:
@Geod24 was right, I was missing a final blocking call in the unittest. Pushed a fix.

Geod24 (approved):

AndrejMitrovic:
It's green!

Geod24:
That's a first :D

PR #11, opened 2019-07-17
Add support for simplified non-overload API filtering

AndrejMitrovic:


source/geod24/LocalRest.d
```diff
@@ -572,11 +576,24 @@ public final class RemoteAPI (API) : API
...
 
-            // ensure it's used with API.method, *not* RemoteAPI.method which
-            // is an override of API.method. Otherwise mangling won't match!
-            enum mangled = getBestMatch!(__traits(getOverloads, API, method_name));
+            // special-case: no other overloads, and parameter list is empty:
+            // just select that one API method
+            alias Overloads = __traits(getOverloads, API, method_name);
+            static if (Overloads.length == 1 && OverloadParams.length == 0)
+            {
+                immutable pretty = method_name ~ Parameters!(Overloads[0]).stringof;
```
AndrejMitrovic, line 586: I found a new bug here. If I do `immutable pretty = format("%s%s", method.name, Parameters!(Overloads[0]).stringof;);`, I get a segfault..

source/geod24/LocalRest.d
```diff
@@ -1041,8 +1060,10 @@ unittest
 
         override size_t fooCount() { return this.foo_count; }
         override size_t fooIntCount() { return this.foo_int_count; }
+        override size_t barCount() { return this.bar_count; }
         override void foo () { ++this.foo_count; }
         override void foo (int) { ++this.foo_int_count; }
+        override void bar (int) { ++this.bar_count; }  // not in any overload set
```
Geod24, line 1066:
```suggestion
        override void bar (int) { ++this.bar_count; }  // not in any overload set
        // This one is part of the overload set of the node, but not of the API
        // It can't be accessed via API and can't be filtered out
        void bar(string) { assert(0); }
```

AndrejMitrovic:
Applied fix.

source/geod24/LocalRest.d
Geod24: You mean `method_name` right ?

AndrejMitrovic: yes sorry

source/geod24/LocalRest.d
```diff
@@ -1088,6 +1129,13 @@ unittest
     static assert(is(typeof(node.filter!(API.foo))));
     static assert(is(typeof(node.filter!(node.foo))));
 
+    // only method in the overload set that takes a parameter,
+    // should still match a call to filter with no parameters
+    static assert(is(typeof(node.filter!(node.bar))));
+
+    // wrong parameters => fail to compile
+    static assert(!is(typeof(node.filter!(node.bar, float))));
```
Geod24, line 1137:
```suggestion
    static assert(!is(typeof(node.filter!(node.bar, float))));
    // Only `API` overload sets are considered
    static assert(!is(typeof(node.filter!(node.bar, string))));    
```

AndrejMitrovic:
Pushed fix.

Geod24 (approved):

AndrejMitrovic:
There was a failure on MacOS, maybe (just maybe) because I missed a synchronous `bar()` call.

AndrejMitrovic:
Yeah that was it. Since you approved it I'll go ahead and merge it.

AndrejMitrovic:
Wait, I can't. 😆

Geod24:
RIP

PR #12, opened 2019-07-26
Implement request timeouts

AndrejMitrovic:
The second commit is me testing stuff to see if it works.

### DIFF
--- a/source/geod24/LocalRest.d
+++ b/source/geod24/LocalRest.d
@@ -81,18 +81,27 @@ private struct FilterAPI
     string pretty_func;
 }
 
+/// Status of a request
+private enum Status
+{
+    /// Request failed
+    Failed,
+
+    /// Request succeeded
+    Success
+}
+
 /// Data sent by the callee back to the caller
 private struct Response
 {
-    /// `true` if the method returned successfully,
-    /// `false` if an `Exception` occured
-    bool success;
+    /// Final status of a request (failed, success, etc)
+    Status status;
     /// In order to support re-entrancy, every request contains an id
     /// which should be copied in the `Response` so the scheduler can
     /// properly dispatch this event
     /// Initialized to `size_t.max` so not setting it crashes the program
     size_t id;
-    /// If `success == true`, the JSON-serialized return value.
+    /// If `status == Status.Success`, the JSON-serialized return value.
     /// Otherwise, it contains `Exception.toString()`.
     string data;
 }
@@ -328,7 +337,7 @@ public final class RemoteAPI (API) : API
                         {
                             // we have to send back a message
                             import std.format;
-                            C.send(cmd.sender, Response(false, cmd.id,
+                            C.send(cmd.sender, Response(Status.Failed, cmd.id,
                                 format("Filtered method '%%s'", filter.pretty_func)));
                             return;
                         }
@@ -339,20 +348,20 @@ public final class RemoteAPI (API) : API
                         {
                             C.send(cmd.sender,
                                 Response(
-                                    true,
+                                    Status.Success,
                                     cmd.id,
                                     node.%1$s(args.args).serializeToJsonString()));
                         }
                         else
                         {
                             node.%1$s(args.args);
-                            C.send(cmd.sender, Response(true, cmd.id));
+                            C.send(cmd.sender, Response(Status.Success, cmd.id));
                         }
                     }
                     catch (Throwable t)
                     {
                         // Our sender expects a response
-                        C.send(cmd.sender, Response(false, cmd.id, t.toString()));
+                        C.send(cmd.sender, Response(Status.Failed, cmd.id, t.toString()));
                     }
 
                     return;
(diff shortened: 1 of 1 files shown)


source/geod24/LocalRest.d
```diff
@@ -87,6 +86,8 @@ private struct Response
     /// `true` if the method returned successfully,
     /// `false` if an `Exception` occured
     bool success;
+    /// `true` if the request timed-out
+    bool timed_out;
```
Geod24: I'd prefer making `success` an `enum` at this point

AndrejMitrovic: Good point.

source/geod24/LocalRest.d
```diff
@@ -628,18 +648,50 @@ public final class RemoteAPI (API) : API
...
+
+                            if (this.timeout == Duration.init)
+                            {
+                                return C.receiveOnly!(Response);
+                            }
+                            else
+                            {
+                                bool received_value;
+                                Response actual_res;
+                                C.receiveTimeout(this.timeout,
```
Geod24: This returns a `bool` which tells you if it timed out

source/geod24/LocalRest.d
```diff
@@ -628,18 +648,50 @@ public final class RemoteAPI (API) : API
...
+                                    {
+                                        received_value = true;
+                                        actual_res = res;
+                                    });
+
+                                if (received_value)
+                                {
+                                    return actual_res;
+                                }
+                                else
```
Geod24: Technically you don't need to branch here (`return` makes `else` implicit)

source/geod24/LocalRest.d
```diff
@@ -276,16 +288,18 @@ public final class RemoteAPI (API) : API
...
+          timeout = (optional) timeout to use with requests
 
         Returns:
           A `RemoteAPI` owning the node reference
 
     ***************************************************************************/
 
-    public static RemoteAPI!(API) spawn (Impl) (CtorParams!Impl args)
+    public static RemoteAPI!(API) spawn (Impl) (CtorParams!Impl args,
+        Duration timeout = Duration.init)
```
Geod24, line 304: I worship whoever implemented default parameter after variadic args

source/geod24/LocalRest.d
```diff
@@ -1176,3 +1228,80 @@ unittest
...
+    import core.thread;
+    import std.exception;
+
+    static interface API
+    {
+        bool checkTimedOut ();
+        size_t sleepFor (long dur);
+    }
+
+    __gshared API api;  // can't pass it through ctor
```
Geod24: And you don't want to. If you want to contact an API, you store the `Tid` and instantiate a new `RemoteNode`.
`grep` for `tbn` to see how it's done.

AndrejMitrovic: Wait I fixed this.. where did it go.

AndrejMitrovic:
I pushed some fixes. And now the unittests succeed (locally), but I don't trust it.

AndrejMitrovic:
I mean I am not 99% sure if this is correct code. It makes sense to me, but you're more familiar with the code.

source/geod24/LocalRest.d
```diff
@@ -148,7 +160,20 @@ private final class LocalScheduler : C.FiberScheduler
...
+            if (!ptr.c.wait(duration))
+            {
+                this.pending.status = Status.Timeout;
+                this.pending.data = "";
+            }
+        }
+        else
+        {
+            ptr.c.wait();
+        }
```
```d
if (duration == Duration.init)
    ptr.c.wait();
else if (!ptr.c.wait(duration))
    this.pending = Response(Status.Timeout, this.pending.id);
```

source/geod24/LocalRest.d
```diff
@@ -1176,3 +1233,77 @@ unittest
...
+    {
+        void check ();
+        size_t sleepFor (long dur);
+    }
+
+    static class Node : API
+    {
+        override void check ()
+        {
+            auto node = RemoteAPI!API.spawn!Node(50.msecs);
```
Geod24: I would not spawn a `Node` from another thread. At the moment `Node`s should be spawned from the main thread.

AndrejMitrovic:
I pushed a separate commit which sometimes gives me a Json error, but it looks like it's a timing bug..

AndrejMitrovic:
```
~/dev/localrest request-timeouts $ test
Deserializing composite type ArgWrapper!() which has no serializable fields.
Serializing composite type ArgWrapper!() which has no serializable fields
All unit tests have been run successfully.

~/dev/localrest request-timeouts $ test
Deserializing composite type ArgWrapper!() which has no serializable fields.
Serializing composite type ArgWrapper!() which has no serializable fields
Program exited with code -11
```

And sometimes a segfault.

source/geod24/LocalRest.d
```diff
@@ -1176,3 +1233,84 @@ unittest
...
+            return 42;
+        }
+    }
+
+    auto node_1 = RemoteAPI!API.spawn!Node();
+    auto node_2 = RemoteAPI!API.spawn!Node();
+    node_tid = node_2.tid;
+    node_1.check();
+}
+
```
Geod24: That's probably the root of all your problems. How could you not see this ?

AndrejMitrovic:
The issue was the timeout was set too low, so it was an earlier assert that failed (which did not expect a timeout to occur).

I've changed the timeout to 500 msecs instead of 50 msecs. It went green once, let's see if it goes green again.

AndrejMitrovic:
@Geod24 green twice in a row. :)

PR #13, opened 2019-07-26
Fix the unittests for request filtering

AndrejMitrovic:
Previously this test was really weird, it was using tasks to make a node spawn an async request **to itself**.

This was rewritten to have two nodes, one filtered, and one which just invokes the other node.


Geod24 (approved):
Wow yeah why did I ever approve this code ? I'd never write something like this /s

PR #14, opened 2019-07-31
Don't use Thread.sleep() when rejecting requests

AndrejMitrovic:
Essentially the fix is:

- Keep track of when to wake up and start responding to messages
- If the drop option wasn't specified, then we have to respond to the messages which were sent during the sleeping stage. But the only way to reliably do that is to manually keep track of the message queue.

This could could be simplified further I think..

### DIFF
--- a/source/geod24/LocalRest.d
+++ b/source/geod24/LocalRest.d
@@ -406,6 +406,20 @@ public final class RemoteAPI (API) : API
 
         FilterAPI filter;
 
+        void handle (T)(T arg)
+        {
+            static if (is(T == Command))
+            {
+                scheduler.spawn(() => handleCommand(arg, node, filter));
+            }
+            else static if (is(T == Response))
+            {
+                scheduler.pending = arg;
+                scheduler.waiting[arg.id].c.notify();
+            }
+            else static assert(0, "Unhandled type: " ~ T.stringof);
+        }
+
         try scheduler.start(() {
                 bool terminated = false;
                 while (!terminated)
@@ -421,12 +435,11 @@ public final class RemoteAPI (API) : API
                             filter = filter_api;
                         },
                         (Response res) {
-                            scheduler.pending = res;
-                            scheduler.waiting[res.id].c.notify();
+                            handle(res);
                         },
                         (Command cmd)
                         {
-                            scheduler.spawn(() => handleCommand(cmd, node, filter));
+                            handle(cmd);
                         });
                 }
                 // Make sure the scheduler is not waiting for polling tasks
--- a/source/geod24/LocalRest.d
+++ b/source/geod24/LocalRest.d
@@ -400,17 +400,51 @@ public final class RemoteAPI (API) : API
 
     private static void spawned (Implementation) (CtorParams!Implementation cargs)
     {
+        import std.datetime.systime : Clock, SysTime;
+        import std.algorithm : each;
+        import std.range;
+
         scope node = new Implementation(cargs);
         scheduler = new LocalScheduler;
         scope exc = new Exception("You should never see this exception - please report a bug");
 
-        FilterAPI filter;
+        // very simple & limited variant, to keep it performant.
+        // should be replaced by a real Variant later
+        static struct Variant
+        {
+            this (Response res) { this.res = res; this.tag = 0; }
+            this (Command cmd) { this.cmd = cmd; this.tag = 1; }
+
+            union
+            {
+                Response res;
+                Command cmd;
+            }
+
+            ubyte tag;
+        }
+
+        // used for controling filtering / sleep
+        struct Control
+        {
+            FilterAPI filter;    // filter specific messages
+            SysTime sleep_until; // sleep until this time
+            bool drop;           // drop messages if sleeping
+        }
+
+        Control control;
+
+        bool isSleeping()
+        {
+            return control.sleep_until != SysTime.init
+                && Clock.currTime < control.sleep_until;
+        }
 
         void handle (T)(T arg)
         {
             static if (is(T == Command))
             {
-                scheduler.spawn(() => handleCommand(arg, node, filter));
+                scheduler.spawn(() => handleCommand(arg, node, control.filter));
             }
             else static if (is(T == Response))
             {
@@ -420,6 +454,10 @@ public final class RemoteAPI (API) : API
             else static assert(0, "Unhandled type: " ~ T.stringof);
         }
 
+        // we need to keep track of messages which were ignored when
+        // node.sleep() was used, and then handle each message in sequence.
+        Variant[] await_msgs;
+
         try scheduler.start(() {
                 bool terminated = false;
                 while (!terminated)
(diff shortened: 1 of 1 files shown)


source/geod24/LocalRest.d
```diff
@@ -400,34 +400,96 @@ public final class RemoteAPI (API) : API
...
+        import std.range;
+        SysTime sleep_until;
+        bool drop_messages;
+
+        bool isSleeping()
+        {
+            return sleep_until != SysTime.init && Clock.currTime < sleep_until;
+        }
+
+        void handle (T) (T arg)
```
AndrejMitrovic: Not atomic, should be moved to its own commit.

AndrejMitrovic:
Split it up into atomic commits.

source/geod24/LocalRest.d
```diff
@@ -421,12 +434,11 @@ public final class RemoteAPI (API) : API
                             filter = filter_api;
                         },
                         (Response res) {
-                            scheduler.pending = res;
-                            scheduler.waiting[res.id].c.notify();
+                            handle(res);
```
Geod24: I think you could just pass `handle!Response, handle!Command` then

AndrejMitrovic: what

AndrejMitrovic: Ah you mean `&handle!Response`

Geod24: Yup, sorry

AndrejMitrovic: I wanted to do it just now, but then I realized in the second commit we add a bit of logic here.

source/geod24/LocalRest.d
```diff
@@ -400,12 +400,23 @@ public final class RemoteAPI (API) : API
...
+        import std.datetime.systime;
+        import std.range;
         scope node = new Implementation(cargs);
         scheduler = new LocalScheduler;
         scope exc = new Exception("You should never see this exception - please report a bug");
 
         FilterAPI filter;
 
+        bool drop_messages;
+        SysTime sleep_until;
```
Geod24: Can we unify those 3 things under one local `struct` ?

source/geod24/LocalRest.d
```diff
@@ -419,27 +430,59 @@ public final class RemoteAPI (API) : API
...
-                            handle(res);
+                            if (!isSleeping())
+                            {
+                                handle(res);
+                            }
+                            else
+                            {
+                                if (!drop_messages)
+                                    response_queue ~= res;
+                            }
```
```d
if (!isSleeping())
    handle(res)
else if (!drop_messages)
    response_queue ~= res;
```

source/geod24/LocalRest.d
```diff
@@ -419,27 +430,59 @@ public final class RemoteAPI (API) : API
...
                         },
                         (Command cmd)
                         {
-                            handle(cmd);
+                            if (!isSleeping())
+                            {
+                                handle(cmd);
+                            }
+                            else
```
Geod24: Same

source/geod24/LocalRest.d
```diff
@@ -419,27 +430,59 @@ public final class RemoteAPI (API) : API
...
                         });
+
+                    // now handle any leftover messages after any sleep() call
+                    if (!isSleeping())
+                    {
+                        response_queue.each!(res => handle(res));
+                        response_queue.length = 0;
+
+                        command_queue.each!(res => handle(res));
+                        command_queue.length = 0;
```
Geod24: There's a problem with this: Messages need to be processed in the order they are received

AndrejMitrovic: That's a good point.

AndrejMitrovic:
So the only way I've made this work is by using Variant. I'm not sure if there is a better way to do it.

source/geod24/LocalRest.d
```diff
@@ -413,21 +452,39 @@ public final class RemoteAPI (API) : API
...
+                        await_msgs.each!((Variant msg)
+                        {
+                            if (auto cmd = msg.peek!Command)
+                                handle(*cmd);
+                            else if (auto res = msg.peek!Response)
+                                handle(*res);
+                            else assert(0, "Unhandled message: %s".format(msg));
                         });
+
+                        await_msgs.length = 0;
```
AndrejMitrovic, line 492:
```suggestion
                        await_msgs.length = 0;
                        assumeSafeAppend(await_msgs);
```

?

source/geod24/LocalRest.d
```diff
@@ -40,6 +40,7 @@ module geod24.LocalRest;
 import vibe.data.json;
 
 static import C = std.concurrency;
+import std.datetime.systime : Clock, SysTime;
```
AndrejMitrovic: Move this where it's used?

AndrejMitrovic:
Ok using a union for the messages. We can use a proper variant later, to prevent pulling in dependencies right now. (Rule of three, amirite?)

source/geod24/LocalRest.d
```diff
@@ -400,11 +400,64 @@ public final class RemoteAPI (API) : API
 
     private static void spawned (Implementation) (CtorParams!Implementation cargs)
     {
+        import std.datetime.systime : Clock, SysTime;
+        import std.algorithm;
```
Geod24: Can you narrow down this import ?

source/geod24/LocalRest.d
```diff
@@ -400,11 +400,64 @@ public final class RemoteAPI (API) : API
...
+            {
+                Response res;
+                Command cmd;
+            }
+
+            ubyte tag;
+        }
+
+        // can't embed SysTime in TimeCommand as it has internal aliasing
+        struct Sleep
```
Geod24: What I had in mind was more of a `Control` struct where `filter` would be stored as well. But not a big deal.

source/geod24/LocalRest.d
```diff
@@ -400,11 +400,64 @@ public final class RemoteAPI (API) : API
...
+                scheduler.spawn(() => handleCommand(arg, node, filter));
+            }
+            else static if (is(T == Response))
+            {
+                scheduler.pending = arg;
+                scheduler.waiting[arg.id].c.notify();
+            }
+        }
+
+        /// messages which are waiting to be handled due to a node.sleep() call
```
Geod24: Extra `/` here

source/geod24/LocalRest.d
```diff
@@ -400,11 +400,64 @@ public final class RemoteAPI (API) : API
...
+        // if the node is "sleeping", but doesn't want to drop messages,
+        // it should store them here. we need to track the queue of messages
```
Geod24: That sentence gave me cancer.
```
Stores messages which are waiting to be handled, for example due to a `node.sleep()` call with `drop = false`.
We need to [...]

File: source/geod24/LocalRest.d [hunk 13]
```diff
@@ -400,11 +400,64 @@ public final class RemoteAPI (API) : API
...
+        // on our own, to avoid accidentally dropping priority messages in
```
Geod24: priority messages ?

Geod24 (approved):
Just a few nits otherwise LVGTM

AndrejMitrovic:
Ok I fixed all that stuff.

Geod24 (approved):
I assume there's not good way to test this, as it is a race condition.

PR #15, opened 2019-08-06
Silence warning when composite type is empty

Geod24:
As you can see from the comment, it was always intended to do this...


AndrejMitrovic:
LGTM. Can't merge. :)

PR #21, opened 2019-09-18
Reimplement FiberScheduler and increase the fiber stack size

AndrejMitrovic:
I hate doing this, but unless https://issues.dlang.org/show_bug.cgi?id=20225 is fixed we can't change the stack size of the fibers.

We could wait until the issue is fixed, but this is needed for interacting with the stack-hungry SCP.

https://github.com/dlang/phobos/pull/7193

### DIFF
--- a/source/geod24/LocalRest.d
+++ b/source/geod24/LocalRest.d
@@ -118,8 +118,180 @@ private struct ArgWrapper (T...)
     T args;
 }
 
+/**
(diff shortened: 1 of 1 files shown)


AndrejMitrovic:
It seems this messes with the coverage report. I'll submit a patch upstream to make the fiber size configurable, but until then we might need this.

Geod24 (approved):
I think we really need to expand on this, but as a first step it's good.

PR #23, opened 2019-10-10
Add support for explicitly shutting down threads.

### LINKED ISSUE #22: There needs to be a clean way to shut down threads
Currently, the only way to shut down a spawned RemoteAPI thread is to kill off the main thread (which will send a `OwnerTerminated` message to its child threads).

But, in Agora we use explicit shutdowns for all of our tests:

https://github.com/bpfkorea/agora/blob/7f6e46fc20f39b9c7d2abe63b53d900619386376/source/agora/test/Base.d#L164

This is insufficient, because the `shutdown()` method there only dumps / deallocates / destroys internal handles, but the main thread associated with each node continues to run (and will continue to listen to events).

There needs to be a way to tell the node (thread) to stop processing events. A simple `ctrl.shutdown()` API could do the job.

AndrejMitrovic:
Fixes #22

### DIFF
--- a/source/geod24/LocalRest.d
+++ b/source/geod24/LocalRest.d
@@ -112,6 +112,11 @@ private struct TimeCommand
     bool drop = false;
 }
 
+/// Ask the node to shut down
+private struct ShutdownCommand
+{
+}
+
 /// Filter out requests before they reach a node
 private struct FilterAPI
 {
@@ -680,6 +685,9 @@ public final class RemoteAPI (API) : API
                 {
                     C.receiveTimeout(10.msecs,
                         (C.OwnerTerminated e) { terminated = true; },
+                        (ShutdownCommand e) {
+                            terminated = true;
+                        },
                         (TimeCommand s)      {
                             control.sleep_until = Clock.currTime + s.dur;
                             control.drop = s.drop;
@@ -785,6 +793,17 @@ public final class RemoteAPI (API) : API
         public C.Tid tid () @nogc pure nothrow
         {
             return this.childTid;
+        }
+
+        /***********************************************************************
+
+            Send an async message to the thread to immediately shut down.
+
+        ***********************************************************************/
+
+        public void shutdown () @trusted
+        {
+            C.send(this.childTid, ShutdownCommand());
         }
 
         /***********************************************************************
@@ -993,6 +1012,7 @@ unittest
 
     scope test = RemoteAPI!API.spawn!MockAPI();
     assert(test.pubkey() == 42);
+    test.ctrl.shutdown();
 }
 
 /// In a real world usage, users will most likely need to use the registry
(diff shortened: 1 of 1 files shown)


source/geod24/LocalRest.d
```diff
@@ -702,7 +719,7 @@ public final class RemoteAPI (API) : API
                         });
 
                     // now handle any leftover messages after any sleep() call
-                    if (!isSleeping())
+                    if (!terminated && !isSleeping())
```
Geod24: Why this clause ?

source/geod24/LocalRest.d
```diff
@@ -787,6 +804,20 @@ public final class RemoteAPI (API) : API
...
+            Send a message to the thread to immediately shut down.
+            This method will block until the child thread shuts down.
+
+        ***********************************************************************/
+
+        public void shutdown () @trusted
+        {
+            import std.exception;
+            C.send(this.childTid, ShutdownCommand(C.thisTid()));
+            enforce(C.receiveTimeout(10.seconds, (ShutdownCommand s) => s));
```
Geod24: That's a looooong timeout. Maybe use the node's timeout when available ?

Geod24 (approved):

source/geod24/LocalRest.d
AndrejMitrovic: Haha, random choice! Good suggestion.

source/geod24/LocalRest.d
AndrejMitrovic: Ah, I think this was left in during testing. I don't think it should be a problem to process the messages first. Let me try to remove it.

AndrejMitrovic:
Updated: https://github.com/Geod24/localrest/compare/1d8a8b97af43e30353be9093f5d5e176246c76ac..3adf27807947154154d0ea1516d91771dbb0ecca

Is the 5-second default timeout ok if one isn't set? I'm not sure if we should block forever or just pick some "ok" value in this case.

Geod24:
Hum, then is it a good thing to wait for an acknowledgment ? Other commands just send a message and don't wait for an acknowledgment. And it's unlikely one will *want* to interact with the node after its shutdown.

AndrejMitrovic:
That probably works, I was just being precautious about timing issues. Let's try it without the timeout first then. Give me one sec..

AndrejMitrovic:
Updated. Tests still pass, and the code is simpler now. I hope this works for Agora too.

source/geod24/LocalRest.d
```diff
@@ -1252,6 +1279,9 @@ unittest
     // (e.g. Travis Mac testers) so be safe
     assert(node.getCounter() >= 9);
     assert(node.getCounter() == 0);
+
+    import std.range;
```
AndrejMitrovic: Ugh, dead import. Fixing..

AndrejMitrovic:
Removed dead import.

PR #24, opened 2019-10-14
Reimplement fiber scheduler to handle timed-out responses

AndrejMitrovic:
This PR now implements fixes for two bugs:

Child node to child node response <> request ID mismatch. (by fixing the FiberScheduler)
Main thread to child node response <> request ID mismatch. (by using the fixed FiberScheduler)

-------

Notice how nasty this bug is, because vibe.d actually converted the integer of `sleepFor` into a floating point.

### DIFF
--- a/source/geod24/LocalRest.d
+++ b/source/geod24/LocalRest.d
@@ -444,6 +444,9 @@ private final class LocalScheduler : BaseFiberScheduler
 /// However, the one in `std.concurrency` is process-global (`__gshared`)
 private LocalScheduler scheduler;
 
+/// Whether this is the main thread
+private bool is_main_thread;
+
 
 /*******************************************************************************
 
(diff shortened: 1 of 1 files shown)


Geod24:
WHAT

AndrejMitrovic:
There is a somewhat simple way to work around this: make the main thread use a request ID (can be a simple counter), and then drop all requests previous to the counter.

I'll attach a fix here in a few minutes.

AndrejMitrovic:
I included a very simple fix which just makes the main thread drop responses which don't have the same ID as the one that was issued in the request.

Geod24:
If we just use a scheduler we gain the ability to use tasks for free(1)


(1) Conditions may apply.

AndrejMitrovic:
I got a little scared trying to use it.

AndrejMitrovic:
Hmm.. well.. it almost works. But I just noticed there is a flaw in our current fiber scheduler implementation in LocalRest.

The scheduler sets request IDs as no longer being busy here:
https://github.com/Geod24/localrest/blob/beb56b2a7ebc0927ffdc348ed3acb517d7ae6b16/source/geod24/LocalRest.d#L389 This is also in case of timeouts. So if a request times-out, the request ID will be reused here: https://github.com/Geod24/localrest/blob/beb56b2a7ebc0927ffdc348ed3acb517d7ae6b16/source/geod24/LocalRest.d#L365-L366 In the meantime, the "timed-out" request may get a delayed response. 

To put this in simplistic terms:

node A: issues a request with a timeout with ID 0
node B: stalls (maybe it was a just thread sleep or a long operation)
node A: request 0 times out, the ID 0 is marked as being "free"
...
node A: issues a second (unrelated) request, the scheduler says "use ID 0, it's free!"
node B: thread sleep finished, it returns back a response with ID 0
node A: receives response with ID 0, but it contains the response of the previous request, not the last one.

I think we need to change the scheduler a little bit. I can think of a few ideas:

1. Use a hashmap to store pending requests (that seems pretty much necessary)
2. Change the ID number generation to either be:
    - a) random
    - b) a hash of the request (the "command")

AndrejMitrovic:
So I pushed two more commits, one changing the scheduler to use unique IDs. And another commit making the main thread use the scheduler. That last commit, I'm not sure about it. It only runs the "event loop" until the wanted request ID receives a response, and then the event loop is finished until the next request is made when the event loop is started again.

AndrejMitrovic:
Actually I think I can just use `yield` instead of starting/shutting down that event loop each time. That should work.

AndrejMitrovic:
This PR now implements fixes for two bugs:

- Child node to child node response <> request ID mismatch. (by fixing the FiberScheduler)
- Main thread to child node response <> request ID mismatch. (by using the fixed FiberScheduler)

source/geod24/LocalRest.d
```diff
@@ -444,6 +444,9 @@ private final class LocalScheduler : BaseFiberScheduler
 /// However, the one in `std.concurrency` is process-global (`__gshared`)
 private LocalScheduler scheduler;
 
+/// The main thread scheduler, used to wait for specific events
+private LocalScheduler main_scheduler;
```
Geod24: Why introduce another scheduler and not use `scheduler` instead ?

AndrejMitrovic: This was just used to "mark" the thread as being the main thread. We can use a boolean instead. I'm pushing a small change.

AndrejMitrovic:
What makes this not ideal is that I can't just run the `receiveTimeout` task indefinitely, because there is no main fiber to yield to. So I just run the `receiveTimeout` task until the message we are interested in is received.

AndrejMitrovic:
But maybe it should work this way. I kind of can't see it working another way. Do you have a better idea?

Geod24 (approved):
LGTM, great job tracking those issues down

PR #25, opened 2019-11-06
Implement an alternative to `std.concurrency`

MichaelKim20:
> The existing code is too good.

`LocalRest` is great.

> The purpose of PR 

The purpose of the project was to process data interchangeable between fiber using the same structure as `Golang's Channel`.

> Added Features

We have built-in `timeout` function.
Save the status so that it is safely handled during 'shutdown'.

> Deleted Features

'priority' message type has been removed.

> `LocalRest`

The node is threaded and the request and response are handled by the fiber.

PR #28, opened 2019-12-10
Import concurrency in LocalRest, split off registry

Geod24:
Better reviewed commit-by-commit.
Should simplify the diff in #25 
I can further break it down into PR if needed.

### DIFF
--- a/source/geod24/Registry.d
+++ b/source/geod24/Registry.d
@@ -1,0 +1,103 @@
+/*******************************************************************************
+
+    Registry implementation for multi-threaded access
+
+    This registry allows to look up a `Tid` based on a `string`.
+    It is extracted from the `std.concurrency` module to make it reusable
+
+*******************************************************************************/
+
+module geod24.Registry;
+
+import core.sync.mutex;
+import geod24.concurrency;
+
+/// Ditto
+public shared struct Registry
+{
+    private Tid[string] tidByName;
+    private string[][Tid] namesByTid;
+    private Mutex registryLock;
+
+    /// Initialize this registry, creating the Mutex
+    public void initialize() @safe nothrow
+    {
+        this.registryLock = new shared Mutex;
+    }
+
+    /**
+     * Gets the Tid associated with name.
+     *
+     * Params:
+     *  name = The name to locate within the registry.
+     *
+     * Returns:
+     *  The associated Tid or Tid.init if name is not registered.
+     */
+    Tid locate(string name)
+    {
+        synchronized (registryLock)
+        {
+            if (shared(Tid)* tid = name in this.tidByName)
+                return *cast(Tid*)tid;
+            return Tid.init;
+        }
+    }
+
+    /**
+     * Associates name with tid.
+     *
+     * Associates name with tid in a process-local map.  When the thread
+     * represented by tid terminates, any names associated with it will be
+     * automatically unregistered.
+     *
+     * Params:
+     *  name = The name to associate with tid.
+     *  tid  = The tid register by name.
+     *
+     * Returns:
+     *  true if the name is available and tid is not known to represent a
+     *  defunct thread.
+     */
+    bool register(string name, Tid tid)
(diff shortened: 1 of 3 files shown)


Geod24:
I have to check, but this might allow us to get rid of this hack: https://github.com/bpfkorea/agora/blob/de09825c1285a77a7fe29d51ce9e2fffcfdee5a6/source/agora/test/Base.d#L544

AndrejMitrovic:
> Copied from Phobos HEAD

For a second I was worried about using untested changes from HEAD, but after [looking](https://github.com/dlang/phobos/commits/master/std/concurrency.d) I've noticed who the last commit author is.

source/geod24/Registry.d
```diff
@@ -0,0 +1,106 @@
...
+
+*******************************************************************************/
+
+module geod24.Registry;
+
+import core.sync.mutex;
+import geod24.concurrency;
+
+/// Ditto
+public shared struct Registry
```
AndrejMitrovic, line 16: What are the semantics of a `shared` struct? I've heard of `synchronized` classes, but never `shared` aggregates.

AndrejMitrovic: Ah let me guess, everything inside is like `shared:` for fields.

source/geod24/Registry.d
```diff
@@ -0,0 +1,106 @@
...
+     *  defunct thread.
+     */
+    bool register(string name, Tid tid)
+    {
+        synchronized (registryLock)
+        {
+            if (name in tidByName)
+                return false;
+            /*
+              TODO: This check was in `std.concurrency` but `isClosed` is private
```
AndrejMitrovic: But it's in `geod24.*` now, you could make it public. ^^

source/geod24/Registry.d
AndrejMitrovic: I wonder if you can declare it as `shared synchronized struct`, I think this would make all methods lock on the built-in mutex. Concurrency keyword soup, haha.

AndrejMitrovic: (but actually don't do that, it's better to be explicit IMO)

source/geod24/concurrency.d
```diff
@@ -1133,7 +1027,6 @@ struct ThreadInfo
             _send(MsgType.linkDead, tid, ident);
         if (owner != Tid.init)
             _send(MsgType.linkDead, owner, ident);
-        unregisterMe(); // clean up registry entries
```
AndrejMitrovic: Can you comment in the commit message why this part was removed?

AndrejMitrovic:
I guess this makes sense if we're going to (gradually or not) modify std.concurrency to our liking. Done with review.

source/geod24/Registry.d
Geod24: You can't have `synchronized` on `struct` anyway, it's only for classes.

source/geod24/Registry.d
Geod24: `package`, right

Geod24:
> For a second I was worried about using untested changes from HEAD, but after looking I've noticed who the last commit author is.

A trustworthy chap we got here.

Geod24:
Dealt with feedback

source/geod24/Registry.d
Geod24: So there is nothing to do here, right ?

source/geod24/concurrency.d
AndrejMitrovic: Any update?

AndrejMitrovic:
LGTM except Travis failing.

source/geod24/concurrency.d
Geod24: I did add it to the commit

AndrejMitrovic (approved):

AndrejMitrovic:
Merge away when ready.

PR #29, opened 2019-12-11
Implement a VirtualClock, and use it in RemoteAPI

AndrejMitrovic:
I went through several different designs, trying to make things work. I tried out Stellar's VirtualClock implementation, but it doesn't seem like a model that can fit to our use-case (we have per-node threads, they have AFAICT only one main and one worker thread and that's it).

The previous designs based on Stellar's VirtualClock were complex and had flaws.

So instead, I've implemented a much simpler design. I thought about what we needed, and then designed around that:

- We need a way to "pause" time for a node. This is not the same as making a node unresponsive via `ctrl.sleep`, because in the `ctrl.sleep` scenario when the node wakes up its monotonic clock would have still made progress, and any events which were waiting will suddenly start firing.

- We need a way to resume paused time (obviously).

- We need a way to time-travel. This is especially useful in some scenarios: if we expect a node to generate some data (reach consensus every 10 minutes), it would be great to not have to wait for 10 minutes to physically pass.

- Time travel also allows us to test time drifting behavior.

-----

### DIFF
--- a/source/geod24/LocalRest.d
+++ b/source/geod24/LocalRest.d
@@ -78,6 +78,8 @@
 
 module geod24.LocalRest;
 
+import geod24.VirtualClock;
+
 import vibe.data.json;
 
 static import C = geod24.concurrency;
@@ -110,6 +112,32 @@ private struct TimeCommand
     Duration dur;
     /// Whether or not affected messages should be dropped
     bool drop = false;
+}
+
+/// Ask the node for the time / pause / resume / offset its virtual clock time
+private struct GetTime
+{
+    /// Tid of the command-issuing thread
+    C.Tid tid;
+
+    /// will contain the time when the structure is returned
+    MonoTime time;
+}
+
+/// Ditto
+private struct PauseTime
+{
+}
+
+/// Ditto
+private struct ResumeTime
+{
+}
+
+/// Ditto
+private struct OffsetTime
+{
+    Duration offset;
 }
 
 /// Ask the node to shut down
@@ -351,6 +379,16 @@ private final class LocalScheduler : BaseFiberScheduler
     /// Request IDs waiting for a response
     private Waiting[ulong] waiting;
 
+    /// Clock
+    private VirtualClock clock;
+
+
+    /// Ctor
+    public this (VirtualClock clock) nothrow @nogc @safe
+    {
+        this.clock = clock;
+    }
+
     /// Should never be called from outside
     public override Condition newCondition(Mutex m = null) nothrow
     {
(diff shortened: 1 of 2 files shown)


AndrejMitrovic:
This seems to fail after https://github.com/Geod24/localrest/pull/28 was merged. I'm looking into it..

Edit: Ok fixed, it was a missing shutdown call.

source/geod24/VirtualClock.d
```diff
@@ -0,0 +1,125 @@
+/*******************************************************************************
+
+    Author:         Andrej Mitrovic
+    License:        MIT (See LICENSE.txt)
+    Copyright:      Copyright (c) 2018-2019 Mathias Lang. All rights reserved.
+
+    Contains an implementation of a virtual clock. The clock by default follows
+    the monotonic clock, but may be paused, resumed, and can even time-travel
+    by settings its time offset.
```
Geod24, line 5: Blast from the past! Also this documentation is part of `Copyright`

source/geod24/VirtualClock.d
```diff
@@ -0,0 +1,125 @@
...
+
+        Constructor
+
+        The virtual clock begins keeping track of time right away.
+
+    ***************************************************************************/
+
+    public this () nothrow @nogc @safe
+    {
+        this.start_time = MonoTime.currTime();
```
Geod24, line 42: Shouldn't you be able to provide a "current time" ?

source/geod24/VirtualClock.d
```diff
@@ -0,0 +1,125 @@
...
+
+        Resume the clock from the last paused time.
+        If the clock was not paused it's a no-op.
+
+    ***************************************************************************/
+
+    public void resume () nothrow @nogc @safe
+    {
+        if (!this.paused)
+            return;
```
Geod24, line 80: Untested branch

source/geod24/LocalRest.d
```diff
@@ -1810,3 +1931,84 @@ unittest
         assert(ex.msg == `"Request timed-out"`);
     }
 }
+
+version (none):
```
Geod24: I guess that's still a WIP ?

Geod24 (commented):
Do you have an example on how Agora will use it ?

source/geod24/VirtualClock.d
AndrejMitrovic: Woops!

AndrejMitrovic:
> Do you have an example on how Agora will use it ?

I'm working on that now.

source/geod24/LocalRest.d
AndrejMitrovic: Woops, that shouldn't have been left there.

source/geod24/VirtualClock.d
AndrejMitrovic: Well, usually I would want a node to "go forwards in time for X minutes" and trigger a consensus round, for example. I'm not sure if we need the exact start time, although I could add it as a feature.

AndrejMitrovic:
This can be closed. We went with an alternate implementation in Agora itself. And actually, I think there were some drawbacks with the PR code.

PR #33, opened 2020-01-14
Remove ability to link threads, make 'Scheduler' private

Geod24:
Looks like there was a bug in the links implementation: Even if `linked` was `false`, it would be added to the AA, which means it would be iterated on when doing `foreach (tid; links.keys)`


Geod24:
This one increases coverage @AndrejMitrovic :D

PR #36, opened 2020-01-16
Remove global scheduler instance and 'setMailBoxSize'

Geod24:
See each commit for the rationale.


AndrejMitrovic (approved):
I see red, I approve.

PR #37, opened 2020-01-29
Correctly implement the shutdown behavior

AndrejMitrovic:
Creating a proper test-case was difficult.

There is actually an additional test-case which is the originally one I wanted to reproduce, but it depends on timing so it's impossible to make it consistently reproducible.

That test-case would be:

- sleep the node
- send a command (e.g. API method called segfault()) => it's queued in `await_msgs` because of the previous sleep call.
- send a shutdown command **exactly** after the sleep timeout finishes but **before** the event loop reaches: https://github.com/Geod24/localrest/blob/7acf11035b7fffd55c6cb794e89363c5ba6c5435/source/geod24/LocalRest.d#L501 It's not possible to reliably trigger this scenario because of the small timeout here: https://github.com/Geod24/localrest/blob/7acf11035b7fffd55c6cb794e89363c5ba6c5435/source/geod24/LocalRest.d#L474 which runs in the event loop.

### DIFF
--- a/source/geod24/LocalRest.d
+++ b/source/geod24/LocalRest.d
@@ -468,14 +468,11 @@ public final class RemoteAPI (API) : API
         Variant[] await_msgs;
 
         try scheduler.start(() {
-                bool terminated = false;
-                while (!terminated)
+                while (1)
                 {
                     C.receiveTimeout(C.thisTid(), 10.msecs,
-                        (C.OwnerTerminated e) { terminated = true; },
-                        (ShutdownCommand e) {
-                            terminated = true;
-                        },
+                        (C.OwnerTerminated e) { throw exc; },
+                        (ShutdownCommand e) { throw exc; },
                         (TimeCommand s)      {
                             control.sleep_until = Clock.currTime + s.dur;
                             control.drop = s.drop;
@@ -505,8 +502,6 @@ public final class RemoteAPI (API) : API
                         assumeSafeAppend(await_msgs);
                     }
                 }
-                // Make sure the scheduler is not waiting for polling tasks
-                throw exc;
             });
         catch (Exception e)
             if (e !is exc)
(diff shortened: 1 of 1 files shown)


source/geod24/LocalRest.d
```diff
@@ -497,8 +499,16 @@ public final class RemoteAPI (API) : API
                                 await_msgs ~= Variant(cmd);
                         });
 
-                    // now handle any leftover messages after any sleep() call
-                    if (!isSleeping())
+                    // if this node was sleeping and then terminated then we cannot process any leftover requests from
+                    // other nodes, however we must notify other nodes that their requests have failed.
```
Geod24: Why do we need to notify of failure ? This is the difference between a "clean" and an immediate shutdown. In a clean shut down you get an RST (or the equivalent for your protocol), but in an immediate shutdown you just time out.
I would think it is more robust to always expose immediate shutdown semantics, no ?

AndrejMitrovic: Right. I was thinking in terms of what is more useful to the programmer, as a debugging tool. But maybe having immediate shutdown semantics makes much more sense.

In that case, the solution is much simpler. We just throw the preallocated exception in the event loop.

AndrejMitrovic:
Updated the PR with a simpler shutdown behavior.

source/geod24/LocalRest.d
```diff
@@ -1594,3 +1589,46 @@ unittest
...
+        }
+    }
+
+    auto node_1 = RemoteAPI!API.spawn!Node();
+    auto node_2 = RemoteAPI!API.spawn!Node();
+    node_tid = node_2.tid;
+    node_1.check();
+    node_2.ctrl.shutdown();  // shut it down before wake-up, segfault() command will be ignored
+
+    Thread.sleep(1.seconds);
```
Geod24: What's the intent of the sleep ? To make sure the Thread terminates ? If so joining the thread might be better no ?

AndrejMitrovic:
Updated!

Geod24 (approved):

Issue #39, opened 2020-02-11
Remove usage of globals in concurrency


Geod24:
There shouldn't be any need for global variable in `concurrency`, so any use should be removed.

PR #43, opened 2020-02-13
Remove usage of globals in concurrency

MichaelKim20:
When calling function `thisInfo`, a global variable is created for each thread.
I made an `InfoThread` that inherited Thread.
It has a `ThreadInfo` inside. 
Therefore, if  `InfoThread` is used, the use of global variables can be reduced.
Issues is #39 Remove usage of globals in concurrency

### DIFF
--- a/source/geod24/concurrency.d
+++ b/source/geod24/concurrency.d
@@ -217,6 +217,11 @@ private
 
     @property ref ThreadInfo thisInfo() nothrow
     {
+        auto t = cast(InfoThread)Thread.getThis();
+
+        if (t !is null)
+            return t.info;
+
         return ThreadInfo.thisInfo;
     }
 }
@@ -485,7 +490,7 @@ if (isSpawnable!(F, T))
     }
 
     // TODO: MessageList and &exec should be shared.
-    auto t = new Thread(&exec);
+    auto t = new InfoThread(&exec);
     t.start();
     return spawnTid;
 }
@@ -994,6 +999,12 @@ class FiberScheduler
 
         if (f !is null)
             return f.info;
+
+        auto t = cast(InfoThread)Thread.getThis();
+
+        if (t !is null)
+            return t.info;
+
         return ThreadInfo.thisInfo;
     }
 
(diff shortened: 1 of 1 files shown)


MichaelKim20:
@Geod24 Ready for review

source/geod24/concurrency.d
```diff
@@ -849,6 +854,51 @@ struct ThreadInfo
...
+/***************************************************************************
+
+    Thread with ThreadInfo,
+    This is implemented to avoid using global variables.
+
+***************************************************************************/
+
+public class InfoThread : Thread
+{
+    public ThreadInfo info;
```
Geod24, line 867: It is never set though ?

MichaelKim20: It's accessed through thisInfo.
Doesn't it have to be initialized?

Geod24: It does have to be initialized.

MichaelKim20: Its initialization was implemented by ThreadScheduler.spawn.

Geod24: `ThreadScheduler.spawn` is never used. In fact, I'm considering removing `ThreadScheduler` completely, there's nothing special about it. `_spawn` is the one that gets used.

MichaelKim20: I see. _spawn already has initialization.

MichaelKim20:
@Geod24 @AndrejMitrovic  Ready for review

source/geod24/concurrency.d
Geod24: Yes it should just be replaced so that the global can be removed. Otherwise this is just dead code.

MichaelKim20: I returned the ThreadScheduler's code to its original form.

MichaelKim20:
````D

struct ThreadInfo
{
...
    static @property ref thisInfo() nothrow
    {
        static ThreadInfo val;
        return val;
    }
...
}
````
Global variables are used here.
When using InfoThread, the global variables were not used.
However, the main threads and other threads use global variables.

Geod24 (approved):

Issue #40, opened 2020-02-11
Implement buffered channels


Geod24:
Related to #25 :

We want an interface similar to go buffered channels, where write are non-handoff on non-full buffer, and reads are non-handoff on non-empty channels.
On reading empty channels and writing full channels, reader/writer competes for a slot.

Example in Go:
- [Textbook](https://play.golang.org/p/3BRCdRnRszb) (from the [tour](https://gobyexample.com/channel-buffering)).
```golang
package main

import "fmt"

func otherFunc(messages chan string) {
    fmt.Println("+ Entering otherFunc()");
    <- messages
    fmt.Println("= Done for the first wait");
    <- messages
    fmt.Println("- Leaving otherFunc()");
}

func main() {
    // To see a different behavior, remove the second argument to `make`
    messages := make(chan string, 1)
    // messages := make(chan string)
    go otherFunc(messages)
    fmt.Println("* Writing to the channel from main")
    messages <- "Hello World"
    fmt.Println("* No hand off happened, channel is now full")
    messages <- "Handing off"
    fmt.Println("* Done with the hand off")
}
```

omerfirmak:
This should be closed since #96 is merged.

PR #44, opened 2020-02-14
Implement buffered channels. 

MichaelKim20:
This is similar to `Go-Lang channel`
It has one buffer.
It does not block when writing. However, when reading, it blocks if there is no data.

There are two kinds of fiber(thread). One reads, and one writes.

**Buffered channel**
When the write-fiber(thread) is working,
it puts the message in the queue.
When the read-fiber(thread) is working,
take the message out of the queue and take it.
If there are no messages in the queue, wait for write-fiber(thread) to put
the messages in the queue.

The issue is #40 Implement buffered channels

### DIFF
--- a/source/geod24/concurrency.d
+++ b/source/geod24/concurrency.d
@@ -42,6 +42,7 @@ import core.atomic;
 import core.sync.condition;
 import core.sync.mutex;
 import core.thread;
+import std.container;
 import std.range.primitives;
 import std.traits;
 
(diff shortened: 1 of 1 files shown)


source/geod24/concurrency.d
```diff
@@ -1563,3 +1564,168 @@ private:
...
 }
+
+
+/*******************************************************************************
+
+    This is similar to go buffered channels
+
+*******************************************************************************/
+
+public class Channel (T)
```
AndrejMitrovic, line 1638: Missing `Params:` section for the `T` parameter.

source/geod24/concurrency.d
```diff
@@ -1563,3 +1564,168 @@ private:
...
+    public bool tryReceive (T *msg)
+    {
+        this.mutex.lock();
+        scope (exit)
+            this.mutex.unlock();
+
+        if (this.closed)
+            return false;
+
+        if (this.queue[].walkLength > 0)
```
AndrejMitrovic: walkLength is O(n). It is better to use the `empty()` property.

source/geod24/concurrency.d
```diff
@@ -1563,3 +1564,168 @@ private:
...
+
+        Return:
+            msg = message to receive
+
+    ***************************************************************************/
+
+    public T receive ()
+    {
+        T msg;
+        while (!tryReceive(&msg)) {
```
```suggestion
        while (!tryReceive(&msg))
        {
```

source/geod24/concurrency.d
```diff
@@ -1563,3 +1564,168 @@ private:
...
+
+        Params:
+            msg = point of message to send
+
+        Return:
+            If successful, return true, otherwise return false
+
+    ***************************************************************************/
+
+    public bool tryReceive (T *msg)
```
```suggestion
    public bool tryReceive (T* msg)
```

source/geod24/concurrency.d
```diff
@@ -1563,3 +1564,168 @@ private:
...
+    /***************************************************************************
+
+        Return closing status
+
+        Return:
+            If channel is closed, return true, otherwise return false
+
+    ***************************************************************************/
+
+    public @property bool isClosed () @safe @nogc pure
```
AndrejMitrovic: Why is this the only method that is `@safe` / `@trusted`? That means other `@safe` code can only use this function in this class.

AndrejMitrovic (changes requested):

source/geod24/concurrency.d
MichaelKim20: Thank you for your review.
I just added a comment. Is this what you wanted?

AndrejMitrovic: Perfect!

MichaelKim20:
@Geod24 @AndrejMitrovic Ready for review

source/geod24/concurrency.d
```diff
@@ -1563,3 +1564,172 @@ private:
...
+        Params:
+            msg = point of message to send
+
+        Return:
+            If successful, return true, otherwise return false
+
+    ***************************************************************************/
+
+    public bool tryReceive (T* msg)
+    {
```
AndrejMitrovic, line 1826: `assert(msg !is null);`

`tryReceive` will segfault if it's called with an empty pointer.

MichaelKim20: I added it

source/geod24/concurrency.d
MichaelKim20: I fixed it

Issue #41, opened 2020-02-11
Implement unbuffered channels


Geod24:
Implement the default channel semantic of Go, unbuffered channels that provide direct hands off semantics.

In essence, when a write / read happens on a channel that is unbuffered, the control is transferred (handed off) directly to another fiber, possibly one waiting for data.

See the order of things in this example:
```golang
package main

import "fmt"

func notBomberman(messages chan int) {
    fmt.Println("+ Pong()");
    <- messages
    fmt.Println("+ Pong()");
    <- messages
    fmt.Println("+ Pong()");
    <- messages
}

func main() {
    messages := make(chan int)
    go notBomberman(messages)
    fmt.Println("* Ping()")
    messages <- 42
    fmt.Println("* Ping()")
    messages <- 42
    fmt.Println("* Ping()")
    messages <- 42
    fmt.Println("* Boom")
}
```

omerfirmak:
This should be closed since #96 is merged.

PR #45, opened 2020-02-14
Implement unbuffered channels

MichaelKim20:
I Implemented the default channel semantic of Go, unbuffered channels that provide direct hands off semantics.

There are two kinds of fiber(thread). One reads, and one writes.

**Unbuffered channel**

When the write-fiber(thread) is working, 
it sends the data to the read-fiber(thread) if it already exists.
Otherwise, the write-fiber(thread) waits for the read-fiber(thread) to come in.

When read-fiber(thread) works, 
it receives data from the write-fiber(thread) if it already exists.
Otherwise, the read-fiber(thread) waits for the write-fiber(thread) to come in.

The issue is #41 Implement unbuffered channels

### DIFF
--- a/source/geod24/concurrency.d
+++ b/source/geod24/concurrency.d
@@ -42,6 +42,7 @@ import core.atomic;
 import core.sync.condition;
 import core.sync.mutex;
 import core.thread;
+import std.container;
 import std.range.primitives;
 import std.traits;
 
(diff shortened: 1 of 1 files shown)


source/geod24/concurrency.d
```diff
@@ -898,6 +924,51 @@ public class InfoThread : Thread
...
 
+/**
+ * An example Scheduler using kernel threads.
+ *
+ * This is an example Scheduler that mirrors the default scheduling behavior
+ * of creating one kernel thread per call to spawn.  It is fully functional
+ * and may be instantiated and used, but is not a necessary part of the
+ * default functioning of this module.
+ */
+class ThreadScheduler
```
Geod24: ?

MichaelKim20: I made a mistake, I corrected it.

source/geod24/concurrency.d
```diff
@@ -1602,7 +1602,32 @@ private:
 
 /*******************************************************************************
 
-    This is similar to go buffered channels
+    This is similar to go channels.
+
+    If you set the size of the queue to 0 in creator,
```
```suggestion
    If you set the size of the queue to 0 in the constructor,
```

source/geod24/concurrency.d
```diff
@@ -1643,14 +1692,56 @@ public class Channel (T)
...
+        {
+            this.queue.insertBack(msg);
+            this.mcount++;
+            this.mutex.unlock();
+            return true;
+        }
 
-        this.queue.insertBack(msg);
-        return true;
+        scope scheduler = thisScheduler;
```
AndrejMitrovic: I would add a comment here:

```
// queue is full or channel is unbuffered
```

source/geod24/concurrency.d
```diff
@@ -1643,14 +1692,56 @@ public class Channel (T)
...
+        if (Fiber.getThis() && scheduler)
```
```suggestion
        if (Fiber.getThis() && scheduler !is null)
```

source/geod24/concurrency.d
```suggestion
        auto scheduler = thisScheduler;
```

Just a note here: `scope` has no meaning when you assign existing values to variables. It only has meaning when you instantiate a new object. For example:

```D
scope obj = new Class();  // destroyed at end of scope
```

This is different to the following:

```D
auto obj = new Class();
scope var = obj;  // *not* destroyed at end of scope
```

source/geod24/concurrency.d
```diff
@@ -1643,14 +1692,56 @@ public class Channel (T)
...
+        {
```
AndrejMitrovic, line 1721: There's duplication in these two blocks. So I suggest rewriting it:

```D
void waitForReader (Mutex mutex, Condition condition)
{
    ChannelContext!T new_context;
    new_context.msg_ptr = null;
    new_context.msg = msg;
    new_context.mutex = null;
    new_context.condition = condition;
    this.sendq.insertBack(new_context);
    this.mutex.unlock();
    new_context.wait();
}

scope scheduler = thisScheduler;
if (Fiber.getThis() && scheduler)
{
    waitForReader(null, scheduler.newCondition(null));
    return true;
}
else
{
    auto mutex = new Mutex();
    waitForReader(mutex, new Condition(mutex));
    return true;
}
```

source/geod24/concurrency.d
```diff
@@ -1665,12 +1756,58 @@ public class Channel (T)
 
     public T receive ()
     {
+        this.mutex.lock();
+
+        if (this.closed)
+        {
+            this.mutex.unlock();
+            assert(0, "Channel is closed.");
```
AndrejMitrovic, line 1762: Very nice. This is how assertions should be used, to catch programmer errors.

source/geod24/concurrency.d
```diff
@@ -1665,12 +1756,58 @@ public class Channel (T)
...
+        }
+
         T msg;
```
AndrejMitrovic: I would prefer if `T msg;` was declared inside of each `if` block separately. Because it will not be re-used. So:

```D
if (!this.sendq.empty)
{
    T msg;
    ...
}

if (!this.queue.empty)
{
    T msg = this.queue.front;
    ...
}
```

And so on..

source/geod24/concurrency.d
```diff
@@ -1665,12 +1756,58 @@ public class Channel (T)
...
+            this.queue.removeFront();
+            this.mcount--;
+            this.mutex.unlock();
+            return msg;
+        }
+
+        scope scheduler = thisScheduler;
+        if (Fiber.getThis() && scheduler)
+        {
+            ChannelContext!T new_context;
```
AndrejMitrovic, line 1790: These two `if` / `else` blocks again have similar code, so I suggest writing a helper function just like the one I suggested in the `send` routine.

source/geod24/concurrency.d
```diff
@@ -1665,12 +1756,58 @@ public class Channel (T)
...
+            msg = context.msg;
+            this.mutex.unlock();
+            context.notify();
+            return msg;
+        }
+
+        if (!this.queue.empty)
+        {
+            msg = this.queue.front;
+            this.queue.removeFront();
```
AndrejMitrovic, line 1780:
```suggestion
            this.queue.removeFront();
            assert(this.mcount > 0, "mcount was zero!");
```

source/geod24/concurrency.d
```diff
@@ -1691,19 +1828,32 @@ public class Channel (T)
...
+            *(msg) = context.msg;
+            this.mutex.unlock();
+            context.notify();
+            return true;
+        }
 
         if (!this.queue.empty)
         {
             *(msg) = this.queue.front;
             this.queue.removeFront();
```
AndrejMitrovic, line 1850: There is a missing `this.mcount--;` decrement here. The tests need to be improved because they didn't catch this error.

```suggestion
            this.queue.removeFront();
            assert(this.mcount > 0, "mcount was zero!");
            this.mcount--;
```

source/geod24/concurrency.d
```diff
@@ -1734,38 +1884,155 @@ public class Channel (T)
```
AndrejMitrovic, line 1889: The method documentation needs to be updated. It doesn't say what closing the channel actually does. And it looks like it notifies all waiting receivers/senders.

AndrejMitrovic (changes requested):

AndrejMitrovic:
It's a bit strange that https://github.com/Geod24/localrest/pull/44 introduces a `Channel` structure (also included as a commit in this PR), but then this PR #45 immediately changes the code introduced in #44 to allow the channel to be buffered or un-buffered.

This kind of atomic split of the Channel implementation would only make sense if the two PRs were introduced with a long time between them, where the first implementation was used by LocalRest/Agora for a while.

So I suggest merging the two commits into one, which says something like "Introduced (un)buffered Channels".

MichaelKim20:
> It's a bit strange that #44 introduces a `Channel` structure (also included as a commit in this PR), but then this PR #45 immediately changes the code introduced in #44 to allow the channel to be buffered or un-buffered.
> 
> This kind of atomic split of the Channel implementation would only make sense if the two PRs were introduced with a long time between them, where the first implementation was used by LocalRest/Agora for a while.
> 
> So I suggest merging the two commits into one, which says something like "Introduced (un)buffered Channels".

@AndrejMitrovic 's opinion is good.
By the way, the two PRs belong to different issues. And I think I'd like to clarify the technical differences between the two and how they're implemented.
@Geod24 What do you think about this?

MichaelKim20:
@Geod24  @AndrejMitrovic  Ready for review

MichaelKim20:
@AndrejMitrovic Thank you for your kind review.

MichaelKim20:
I merged the two commits into one,
PR #44 and PR #45,
A message of the new commit is "Implement (un)buffered Channel"
@Geod24 @AndrejMitrovic  Ready for review.

source/geod24/concurrency.d
```diff
@@ -1603,3 +1604,449 @@ private:
...
+    ***************************************************************************/
+
+    public void close ()
+    {
+        ChannelContext!T context;
+
+        this.mutex.lock();
+        scope (exit) this.mutex.unlock();
+
+        this.closed = true;
```
AndrejMitrovic, line 1898: Does it make sense to call `close` on a channel multiple times? Because if not then there should be another assert here:

```suggestion
        assert(!this.closed, "Channel was already closed!");
        this.closed = true;
```

source/geod24/concurrency.d
```diff
@@ -1603,3 +1604,449 @@ private:
...
+
+        while (true)
```
```suggestion
        while (!this.recvq.empty)
```

and then you can get rid of the `if / break` below.

source/geod24/concurrency.d
```diff
@@ -1603,3 +1604,449 @@ private:
...
+
+            context = this.recvq.front;
+            this.recvq.removeFront();
+            context.notify();
+        }
+
+        this.queue.clear();
+        this.mcount = 0;
+
+        while (true)
```
```suggestion
        while (!this.sendq.empty)
```

same comment as above.

source/geod24/concurrency.d
```diff
@@ -1603,3 +1604,449 @@ private:
...
+    It has information to use in standby.
+
+***************************************************************************/
+
+private struct ChannelContext (T)
+{
+    /// This is a message. Used in put
+    public T  msg;
+
+    /// This is a message point. Used in get
```
```suggestion
    /// This is a message pointer. Used in get
```

source/geod24/concurrency.d
```diff
@@ -824,6 +824,9 @@ struct ThreadInfo
     Tid ident;
     Tid owner;
 
+    /// Storage of information required for scheduling, message passing, etc.
+    public Object[string] objects;
```
AndrejMitrovic: I am not sure why this is here. This can just be `FiberScheduler scheduler;`, and `thisScheduler` could return it. Why is it a map, and why does it store `Object` instead of `FiberScheduler`?

It is not used to store any other types of objects in this PR.

AndrejMitrovic (changes requested):

source/geod24/concurrency.d
MichaelKim20: All Commitments can be found at PR #25.
It will be used later. What should I do in this case?

https://github.com/Geod24/localrest/pull/25/commits/297220c01b9530b3b12513422d1b05b7f1781222#diff-b2ce8db5a10ee88e89a581259145e669R115-R147

AndrejMitrovic: Alright, so you want to store multiple types into `ThreadInfo`. But why not just store:

```
struct ThreadInfo
{
    FiberScheduler scheduler;
    MessageChannel channel;
}
```

And then add more things as it's necessary? It doesn't have to use a hashmap and be dynamic.

AndrejMitrovic: Ok Michael explained it to me IRL, so the issue is that `ThreadInfo` is in `geod24.concurrency`, but the real type we want to store there is `Channel!Type`, and we don't know the type yet.

There might be a better way to do this, but I'll have to think about it. In the meantime, we could go with Michael's approach.

MichaelKim20: Thank you for your review.

I made the following corrections.
struct ThreadInfo
{
    FiberScheduler scheduler;
}
If I need an Object later, I will add a full explanation there.

source/geod24/concurrency.d
```diff
@@ -2040,3 +2040,331 @@ unittest
...
+version (unittest) void spawnThread (void delegate() op)
+{
+    auto t1 = new InfoThread({
+        thisScheduler = new FiberScheduler();
+        op();
+    });
+    t1.start();
+}
+
+/// Min Thread -> [ channel1 ] -> Thread1 -> [ channel2 ] -> Min Thread
```
```suggestion
/// Parent thread and child thread communicating with 2 channels
```

source/geod24/concurrency.d
```diff
@@ -2040,3 +2040,331 @@ unittest
...
+        channel2.send(msg*msg);
+    });
+
+    // Main Thread1
+    channel1.send(2);
+    assert(channel2.receive() == 4);
+}
+
+
+/// Min Thread -> [ channel1 ] -> Fiber1 in Thread1 -> [ channel2 ] -> Min Thread
```
```suggestion
/// Parent thread and fiber of child thread communicating with 2 channels
```

source/geod24/concurrency.d
```diff
@@ -2040,3 +2040,331 @@ unittest
...
+            channel2.send(msg*msg);
+        });
+    });
+
+    // Main Thread
+    channel1.send(2);
+    assert(channel2.receive() == 4);
+}
+
+/// Fiber in Min Thread -> [ channel1 ] -> Fiber1 in Thread1 -> [ channel2 ] -> Fiber in Min Thread
```
```suggestion
/// Parent thread's fiber and child thread communicating with 2 channels
```

source/geod24/concurrency.d
```diff
@@ -2040,3 +2040,331 @@ unittest
...
+    // Main Thread
+    thisScheduler = new FiberScheduler();
+    thisScheduler.start({
+        channel1.send(2);
+        result = channel2.receive();
+    });
+    assert(result == 4);
+}
+
+/// Fiber1 -> [ channel2 ] -> Fiber2 -> [ channel1 ] -> Fiber1
```
```suggestion
/// 2 fibers in a child thread communicating with each other using 2 channels
```

source/geod24/concurrency.d
```diff
@@ -2040,3 +2040,331 @@ unittest
...
+    synchronized (mutex)
+    {
+        condition.wait(1000.msecs);
+    }
+
+    assert(result == 4);
+}
+
+
+/// Fiber1 in Thread1 -> [ channel2 ] -> Fiber2 in Thread2 -> [ channel1 ] -> Fiber1 in Thread1
```
```suggestion
/// Fiber in parent thread communicating with fiber in child thread using 2 channels
```

source/geod24/concurrency.d
```diff
@@ -2040,3 +2040,331 @@ unittest
...
+
+    synchronized (mutex)
+    {
+        condition.wait(1000.msecs);
+    }
+    assert(result == 4);
+}
+
+
+/// Thread1 -> [ channel2 ] -> Thread2 -> [ channel1 ] -> Thread1
```
```suggestion
/// Two child threads communicating with each other with 2 channels
```

source/geod24/concurrency.d
```diff
@@ -2040,3 +2040,331 @@ unittest
...
+    synchronized (mutex)
+    {
+        condition.wait(3000.msecs);
+    }
+
+    assert(result == 4);
+}
+
+
+/// Thread1 -> [ channel2 ] -> Fiber1 in Thread 2 -> [ channel1 ] -> Thread1
```
```suggestion
/// One child thread communicating with another child thread's fiber using 2 channels
```

source/geod24/concurrency.d
```diff
@@ -2040,3 +2040,331 @@ unittest
...
+unittest
+{
+    auto channel1 = new Channel!int(0);
+    auto channel2 = new Channel!int(1);
+    int result = 0;
+
+    Mutex mutex = new Mutex;
+    Condition condition = new Condition(mutex);
+
+    // Thread1 - It'll be tangled.
```
AndrejMitrovic: The word we use in concurrency is `blocked`, not `tangled`.

source/geod24/concurrency.d
```diff
@@ -2040,3 +2040,331 @@ unittest
...
+        }
+    });
+
+    synchronized (mutex)
+    {
+        condition.wait(1000.msecs);
+    }
+    assert(result == 0);
+
+    // Thread2 - Unravel a tangle
```
```suggestion
    // Thread2 - Unblock a thread
```

source/geod24/concurrency.d
```diff
@@ -2040,3 +2040,331 @@ unittest
...
+    });
+
+    synchronized (mutex)
+    {
+        condition.wait(1000.msecs);
+    }
+    assert(result == 2);
+
+    result = 0;
+    // Thread3 - It'll not be tangled, because queue size is 1
```
```suggestion
    // Thread3 - It'll not be blocked, because queue size is 1
```

source/geod24/concurrency.d
```diff
@@ -2040,3 +2040,331 @@ unittest
...
+
+    synchronized (mutex)
+    {
+        condition.wait(1000.msecs);
+    }
+    assert(result == 4);
+}
+
+
+// If the queue size is 0, it will block when it is sent and received on the same thread.
```
```suggestion
// Test blocking behavior on threads when using unbuffered channels
```

source/geod24/concurrency.d
```diff
@@ -2040,3 +2040,331 @@ unittest
...
+
+    synchronized (mutex)
+    {
+        condition.wait(1000.msecs);
+    }
+    assert(result == 2);
+}
+
+
+// If the queue size is 0, it will block when it is sent and received on the same fiber.
```
```suggestion
// Test blocking behavior on fibers when using unbuffered channels
```

source/geod24/concurrency.d
```diff
@@ -2040,3 +2040,331 @@ unittest
...
+            //  Fiber2 - Unravel a tangle
+            scheduler.spawn({
+                result = channel1.receive();
+                channel1.send(2);
+            });
+
+            cond.wait(1000.msecs);
+            assert(result == 2);
+
+            //  Fiber3 - It'll not be tangled, because queue size is 1
```
```suggestion
            //  Fiber3 - It'll not be blocked, because queue size is 1
```

source/geod24/concurrency.d
```diff
@@ -2040,3 +2040,331 @@ unittest
...
+            scheduler.spawn({
+                channel1.send(2);
+                result = channel1.receive();
+                cond.notify();
+            });
+
+            assert(!cond.wait(1000.msecs));
+            assert(result == 0);
+
+            //  Fiber2 - Unravel a tangle
```
```suggestion
            //  Fiber2 - Unblock a fiber
```

AndrejMitrovic (changes requested):

AndrejMitrovic:
The tests so far look good, however they are incomplete.

There are missing tests for:
- Return value of `send`. They are ignored in all the test-cases.
- There also need to be tests that cause `send` to return false.
- Test calling close()
- Test calling send after calling close()
- Test calling `tryReceive` after calling close(). Unlike `receive` which will `assert` if trying to read on a closed channel, `tryReceive` will return false. It needs a test.
- isClosed() test

MichaelKim20:
@AndrejMitrovic Thank you for your review

MichaelKim20:
> The tests so far look good, however they are incomplete.
> 
> There are missing tests for:
> 
> * Return value of `send`. They are ignored in all the test-cases.
> * There also need to be tests that cause `send` to return false.
> * Test calling close()
> * Test calling send after calling close()
> * Test calling `tryReceive` after calling close(). Unlike `receive` which will `assert` if trying to read on a closed channel, `tryReceive` will return false. It needs a test.
> * isClosed() test

@AndrejMitrovic I checked 'std.concuracy.send'.
There was no return value, and I confirmed using 'assert' inside.
Many people will find it inconvenient to use 'Channel.send' because they have to check the return value every time they use it.
So I changed the return of 'Channel.send' to void.
And I added the test code before and after closing the channel.

MichaelKim20:
@AndrejMitrovic @Geod24 Ready for review

source/geod24/concurrency.d
```diff
@@ -1572,3 +1599,800 @@ private:
...
+
+
+    /***************************************************************************
+
+        Send data `msg`.
+
+        Params:
+            msg = message to send
+
+        Return:
```
AndrejMitrovic: Need to remove `Return` part.

AndrejMitrovic (approved):
Just one small thing left.

AndrejMitrovic:
> @AndrejMitrovic I checked 'std.concurency.send'.
There was no return value, and I confirmed using 'assert' inside.
Many people will find it inconvenient to use 'Channel.send' because they have to check the return value every time they use it.
So I changed the return of 'Channel.send' to void.
And I added the test code before and after closing the channel.

There is one issue: `std.concurrency.send` will not fail to send a request. In the Channel implementation, sending to a closed channel will fail, right? So, you would use code like this:

```D
if (!channel.isClosed())
    channel.send(...);
```

But there is a problem here. There is a race condition. After `isClosed` returns, the mutex is released, and then another thread can call `channel.close`. Meanwhile the main thread will think the channel is not closed, it will call `send()`, and it will `assert(0)` in the body.

So I think we should keep the return value of `send`.

MichaelKim20:
> > @AndrejMitrovic I checked 'std.concurency.send'.
> > There was no return value, and I confirmed using 'assert' inside.
> > Many people will find it inconvenient to use 'Channel.send' because they have to check the return value every time they use it.
> > So I changed the return of 'Channel.send' to void.
> > And I added the test code before and after closing the channel.
> 
> There is one issue: `std.concurrency.send` will not fail to send a request. In the Channel implementation, sending to a closed channel will fail, right? So, you would use code like this:
> 
> ```d
> if (!channel.isClosed())
>     channel.send(...);
> ```
> 
> But there is a problem here. There is a race condition. After `isClosed` returns, the mutex is released, and then another thread can call `channel.close`. Meanwhile the main thread will think the channel is not closed, it will call `send()`, and it will `assert(0)` in the body.
> 
> So I think we should keep the return value of `send`.

I'll think more about this. Thank you.

AndrejMitrovic:
@MukeunKim actually your code is ok. I've had a look at how Go channels are used, and it's the writer thread that ever calls `close` on a channel.

In the future, it would be good to add the thread owner ID to the Channel, so if the wrong thread calls `.close()` on a channel then it would assert.

But for now, your code is good!

MichaelKim20:
@Geod24 Ready for review

Issue #48, opened 2020-02-18
Implement  a reliable `FiberScheduler` in multi-threaded environments


MichaelKim20:
For std.concurrency only one FiberSchedules can be used.
When having one FiberSchedules per thread, the following items are required for stable scheduling.

- [ ] Prevent reentry into the dispatch function.
- [ ] When the thread terminates, it provides a way for the scheduling to terminate reliably.
- [ ] `ThisInfo` is accessible from external modules.

PR #50, opened 2020-02-18
Implement  a reliable `FiberScheduler` in multi-threaded environments

MichaelKim20:
‘start()’ calls during ‘dispatch()’ become ‘assert-failure’.
When the thread terminates, it provides a way for the scheduling to terminate reliably.
Change to `public` visibility attribute of thisInfo

The issue is #48 
This depends on PR #45

### DIFF
--- a/source/geod24/concurrency.d
+++ b/source/geod24/concurrency.d
@@ -42,6 +42,7 @@ import core.atomic;
 import core.sync.condition;
 import core.sync.mutex;
 import core.thread;
+import std.container;
 import std.range.primitives;
 import std.traits;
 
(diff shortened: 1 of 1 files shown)


Geod24:
I'm not sure I understand the need for `SchedulingTerminated` ? It is not thrown anywhere.

MichaelKim20:
> I'm not sure I understand the need for `SchedulingTerminated` ? It is not thrown anywhere.

The main task accepts the request through the channel, and the subtask is created to handle this request. These tasks are managed by the scheduler.
When receiving the end message through the channel, the main task must end scheduling immediately.
The best way to notify the scheduler of a forced termination order is to throw an exception.
So I defined `SchedulingTerminated`, which enabled the Scheduler to end.
Should I add a test code for this? Or do I add this commit when I apply LocalRest?

MichaelKim20:
I added a test code for `SchedulingTerminated`.
I wanted to show you why `SchedulingTerminated` is needed.

@Geod24 Ready for review

Issue #49, opened 2020-02-18
Adapt LocalRest to the new concurrency


MichaelKim20:
The new concurrency has got Channel, FiberScheduler for Multi-Thread.
Adapt LocalRest to the new concurrency using Channel, FiberScheduler for Multi-Thread.

PR #51, opened 2020-02-21
Adapt LocalRest to the new concurrency

MichaelKim20:
The new concurrency has got Channel, FiberScheduler for Multi-Thread.
I adapted LocalRest to the new concurrency using Channel, FiberScheduler for Multi-Thread.

The issue is #49

### DIFF
--- a/source/geod24/LocalRestMessage.d
+++ b/source/geod24/LocalRestMessage.d
@@ -1,0 +1,147 @@
+/*******************************************************************************
+
+    There are various data that nodes need to exchange messages.
+    Send and receive data requests, responses, commands, etc.
+
+*******************************************************************************/
+
+module geod24.LocalRestMessage;
+
+import geod24.concurrency;
+
+import core.time;
+
+public alias MessageChannel = Channel!Message;
+public immutable size_t DefaultQueueSize = 1024;
+
+/// Data sent by the caller
+public struct Command
+{
+    /// Tid of the sender thread
+    Tid sender;
+    /// In order to support re-entrancy, every request contains an id
+    /// which should be copied in the `Response`
+    /// Initialized to `size_t.max` so not setting it crashes the program
+    size_t id = size_t.max;
+    /// Method to call
+    string method;
+    /// Arguments to the method, JSON formatted
+    string args;
+}
+
+/// Ask the node to exhibit a certain behavior for a given time
+public struct TimeCommand
+{
+    /// For how long our remote node apply this behavior
+    Duration dur;
+    /// Whether or not affected messages should be dropped
+    bool drop = false;
+}
+
+/// Ask the node to shut down
+public struct ShutdownCommand
+{
+}
+
+/// Filter out requests before they reach a node
+public struct FilterAPI
+{
+    /// the mangled symbol name of the function to filter
+    string func_mangleof;
+
+    /// used for debugging
+    string pretty_func;
+}
+
+/// Status of a request
+public enum Status
+{
+    /// Request failed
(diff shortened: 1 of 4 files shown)


source/geod24/concurrency.d
```diff
@@ -1105,9 +1105,12 @@ private:
         while (m_fibers.length > 0)
         {
             auto t = m_fibers[m_pos].call(Fiber.Rethrow.no);
-            if (t !is null && !(cast(OwnerTerminated) t))
+            if (t !is null)
             {
-                throw t;
+                if (cast(SchedulingTerminated) t)
+                    break;
+                else if (!(cast(OwnerTerminated) t))
```
AndrejMitrovic, line 1126: I think you can also write it as `else if (cast(OwnerTerminated)t !is null)`. But both ways are fine I think.

source/geod24/concurrency.d
```diff
@@ -1126,6 +1129,143 @@ private:
     size_t m_pos;
 }
 
+/// Test for the use `SchedulingTerminated`
+unittest
+{
+    import std.conv;
+
+    alias ChannelT = Channel!Message;
```
AndrejMitrovic: `Message` is `private`, so it doesn't make sense that this unittest has `///`.

`///` means it will appear in the generated HTML documentation, and `//` means it will not.

source/geod24/concurrency.d
```diff
@@ -1102,6 +1104,11 @@ private:
     {
         import std.algorithm.mutation : remove;
 
+        assert(!this.dispatching, "Already called start. Scheduling already started.");
```
```suggestion
        assert(!this.dispatching, "Already called start(). Scheduling already started.");
```

source/geod24/concurrency.d
```diff
@@ -945,6 +945,8 @@ public class InfoThread : Thread
  */
 class FiberScheduler
 {
+    private bool dispatching;
```
AndrejMitrovic, line 955:
```suggestion
    /// Whether start() has been called
    private bool dispatching;
```

source/geod24/LocalRestMessage.d
```diff
@@ -17,8 +17,8 @@ public immutable size_t DefaultQueueSize = 1024;
 /// Data sent by the caller
 public struct Command
 {
-    /// Tid of the sender thread (cannot be JSON serialized)
-    Tid sender;
+    /// MessageChannel of the sender thread (cannot be JSON serialized)
```
AndrejMitrovic: Is the comment about "cannot be JSON serialized" still true?

AndrejMitrovic (changes requested):

AndrejMitrovic:
@MukeunKim Is this the first part of adapting to Channels? Because there is still the use of globals, for example through the Registry.

MukeunKim:
> @MukeunKim Is this the first part of adapting to Channels? Because there is still the use of globals, for example through the Registry.

Thank you for your review.

I have to do about two more PR in the future.
Everything is in PR #25. The remaining two will be split and PR.
One is to remove features that are not used in the `concurrency.d`.
Another is the implementation of MessagePipeLine.

MichaelKim20:
> @MukeunKim Is this the first part of adapting to Channels? Because there is still the use of globals, for example through the Registry.

@AndrejMitrovic Thank you for your review.

I have to do about two more PR in the future.
Everything is in PR #25. The remaining two will be split and PR.
One is to remove features that are not used in the concurrency.d.
Another is the implementation of MessagePipeLine.
I'll try everything today.

PR #52, opened 2020-02-21
Removed what is not used and applied the style

MichaelKim20:
**Removed what is not used in `concurrency.d`**
  - Tid, Message, MessageBox, Exceptions

**Removed unused imports**

**Applied the style**

The issue is #49


MichaelKim20 (commented):
I couldn't find any corrections.

PR #59, opened 2020-03-19
Implement customizable serialization, make vibe-d:data optional

Geod24:
This will be used by Agora, for types that are not JSON-serializable.

This is marked as a draft ATM, because it doesn't pass the tests with Agora.

### DIFF
--- a/source/geod24/LocalRest.d
+++ b/source/geod24/LocalRest.d
@@ -69,6 +69,10 @@
     the only dependency to Vibe.d is actually to it's JSON module,
     as Vibe.d is the only available JSON module known to the author
     to provide an interface to deserialize composite types.
+    This dependency is however not enforced by the dub project file,
+    as users can provide their own serializer (see `geod24.Serialization`).
+    If the default parameter for serialization is never used,
+    one's project need not depend on `vibe-d:data`.
 
     Author:         Mathias 'Geod24' Lang
     License:        MIT (See LICENSE.txt)
@@ -78,9 +82,8 @@
 
 module geod24.LocalRest;
 
-import vibe.data.json;
-
 static import C = geod24.concurrency;
+import geod24.Serialization;
 import std.meta : AliasSeq;
 import std.traits : Parameters, ReturnType;
 
@@ -99,8 +102,8 @@ private struct Command
     size_t id = size_t.max;
     /// Method to call
     string method;
-    /// Arguments to the method, JSON formatted
-    string args;
+    /// Serialized arguments to the method
+    SerializedData args;
 }
 
 /// Ask the node to exhibit a certain behavior for a given time
@@ -150,9 +153,9 @@ private struct Response
     /// properly dispatch this event
     /// Initialized to `size_t.max` so not setting it crashes the program
     size_t id;
-    /// If `status == Status.Success`, the JSON-serialized return value.
+    /// If `status == Status.Success`, the serialized return value.
     /// Otherwise, it contains `Exception.toString()`.
-    string data;
+    SerializedData data;
 }
 
 /// Simple wrapper to deal with tuples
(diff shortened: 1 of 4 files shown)


AndrejMitrovic:
This is nice. What's the failure in Agora?

Geod24:
Memory corruption because things are moved

Geod24:
This is ready

AndrejMitrovic:
> core.exception.AssertError@source/geod24/concurrency.d(1136): Error: MessageBox is closed

Strange random failure on ldc. I wonder if we have a bug somewhere..

source/geod24/LocalRest.d
```diff
@@ -274,13 +277,23 @@ public void sleep (Duration timeout)
...
     Params:
       API = The interface defining the API to implement
+      Serializer = An aggregate which follows the requirement explained above.
 
 *******************************************************************************/
 
-public final class RemoteAPI (API) : API
+public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
 {
+    static assert (!serializerInvalidReason!(S).length, serializerInvalidReason!S);
```
AndrejMitrovic, line 294: Cute use of the "invalidReason" naming.

Geod24:
Hum, I see it in Agora when I switch to the Serializer. Will look deeper into it.

source/geod24/Serialization.d
```diff
@@ -0,0 +1,259 @@
...
+module geod24.Serialization;
+
+import std.traits;
+
+/*******************************************************************************
+
+    Serialize arbitrary data to / from JSON
+
+    This is the default serializer used by LocalRest, when none is provided.
+    It is a template so that LocalRest does not import Vibe.d if not used.
```
AndrejMitrovic, line 52: Won't it be imported regardless when `-unittest` is passed in the client code? Because the unittest block below will instantiate this template.

Geod24: Unittests in non-root modules should not be parsed

source/geod24/Serialization.d
```diff
@@ -0,0 +1,259 @@
...
+    static if (!is(typeof(() { alias X = S.serialize; })))
+        return "`" ~ S.stringof ~ "` is missing a `serialize` method";
+    else static if (!is(typeof(() { alias X = S.deserialize; })))
+        return "`" ~ S.stringof ~ "` is missing a `deserialize` method";
+
+    /// We need a type to deserialize to test this function
+    /// While we should only test with return types / parameter types,
+    /// pretty much anything that can't (de)serialize an `int` is broken.
+
+    else static if (!is(typeof(() { S.serialize(0); })))
```
AndrejMitrovic, line 95: If `serialize` takes a `ref` this won't work, right? Should we relax that restriction?

source/geod24/Serialization.d
```diff
@@ -0,0 +1,259 @@
...
+        return "`" ~ S.stringof ~ ".serialize` return value should be "
+        ~ "`string`, `immutable(ubyte)[]`, or `immutable(void)[]`, not: `"
+        ~ typeof(S.serialize(0)).stringof ~"`";
+
+    else
+    {
+        // Actual return type used
+        alias RT = SerializedT!S;
+
+        static if (!is(typeof(S.deserialize!int(RT.init))))
```
AndrejMitrovic, line 115: Same question about lvalues.

source/geod24/Serialization.d
Geod24: Hum, good point. Personally I think `serialize` should just take `auto ref` (it has to be a template anyway) if it wants to optimize stuff. Otherwise we might end up in a situation where we inadvertently use lvalues here and it breaks in client code.

Of course all of this will be easier once the rvalueref thing is fixed...

source/geod24/Serialization.d
Geod24: I don't see a reason to take it by `ref` ? It's a const array.

source/geod24/LocalRest.d
```diff
@@ -1619,3 +1634,59 @@ unittest
...
     thread_joinAll();
 }
+
+/// Example of a custom (de)serialization policy
+unittest
+{
+    static struct Serialize
+    {
+    static:
+        public immutable(ubyte[]) serialize (T) (auto ref T value) @trusted
```
AndrejMitrovic, line 1642: Ah I see, lvalues will work but only if `serialize` is a template with `auto ref`.

source/geod24/Serialization.d
AndrejMitrovic: Never mind that question, it's an array.

AndrejMitrovic (approved):
LGTM. Just had a few questions.

source/geod24/Serialization.d
AndrejMitrovic: hmm.. this is when using dub only, right?

source/geod24/Serialization.d
AndrejMitrovic: Ah it has to be templated. Yes that makes sense, I forgot we're dealing with multiple types, doh.

source/geod24/Serialization.d
Geod24: No it's a DMD behavior, see discussions here: https://github.com/dlang/dmd/pull/10937

AndrejMitrovic: Ah interesting. I know Walter implemented the "don't parse unittest blocks when -unittest is not passed", but I didn't know about this other feature.

PR #60, opened 2020-04-23
Throw an Exception type when a method isn't found

TrustHenry:
There was a need to catch exceptions when we cannot find the requested method.
Relates https://github.com/bpfkorea/agora/pull/756

### DIFF
--- a/source/geod24/Exception.d
+++ b/source/geod24/Exception.d
@@ -1,0 +1,17 @@
+/*******************************************************************************
+
+    Defines the exception class thrown when an unimplemented API method
+    is called, or one with the wrong parameters is attempted to be called.
+
+*******************************************************************************/
+
+module geod24.Exception;
+
+import std.exception;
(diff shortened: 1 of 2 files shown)


AndrejMitrovic:
This will not actually be used this way. Because in agora we're always using the most derived API (Validator), so the code in this PR will never be triggered.

We can use an alternate approach. Let's do pair programming again and figure it out.

Geod24 (changes requested):
A few things:
- We don't need a whole module for this. Just put it in `LocalRest.d`;
- No change should be made without a test;
- This behavior should be documented;
- Don't comment out code that is replaced. We have a version control system, so let's use it correctly.

AndrejMitrovic:
This isn't needed anymore, as discussed with @TrustHenry.

TrustHenry:
Yes, I agree.

Issue #64, opened 2020-05-06
Closed mailboxes cannot be handled by client code


Geod24:
Found when investigating this issue with Agora: https://github.com/bpfkorea/agora/issues/796

# Context

A few months ago we unintentionally (and carelessly, to be fair) did a subtle change: we transitioned from silently ignoring `put` on closed message boxes to a hard failure (`assert`). See https://github.com/Geod24/localrest/commit/814377a268953a06c2f1237fc16d3195cb83f80f#diff-7e552a35c6e12cbbfb124195a3e45cd2R1126-R1127
This is something we *do* want, as we're transitioning from message boxes to channels. In the channel model, calling `put` on a closed channel leads to a hard failure (`panic` in Go).

# Today's bug

As I was trying to upgrade LocalRest, the test-suite started to fail. I quickly tracked it down to the aforementioned commit, then spent a few more hours understanding what was going on.

In Agora we have a few asynchronous tasks that perform recurrent actions on timers (e.g. catching-up).
Those tasks currently live independently, and cannot be stopped. In normal mode, they are just run by Vibe.d, while in LocalRest, they are a fiber that gets scheduled periodically.

This is all nice until one tries to shut down the node. Because they do periodic send, they will hit the assert (no one told them the connection was closed, heh), and that'll lead to weird crashes, because threads. I originally thought we needed to stop those tasks, but it's a client-side solution for a library-side problem, and a very un-elegant one.

# Tentative solution

My original approach was simply to add handling code in the sender code to simulate a timeout (which is what one would most likely see in the real world). Spoiler alert: It didn't work. Since it's a race condition. *sometimes* the task that dies is the one that does the request, so when the node respond, *poof*, closed message box.

This then led to another realization: doing the following is subject to the same race condition:
```D
if (!tid.mbox.isClosed())
    send(tid, ...);
```
Because essentially, the message box could have been closed between the check and the send.
This means we need to provide, at the `MessageBox` level, a way to ignore failures, as we need the check to happen after the lock is taken.

Geod24:
Fixed by #65

PR #65, opened 2020-05-09
Properly report internal errors & fix issue 64

Geod24:
See each commit individually & https://github.com/Geod24/localrest/issues/64 for details.
Note that this means we will now provide a big message if a node constructor fails, rendering things like https://github.com/bpfkorea/agora/blob/766e3239f26f161c9475f8b7355f792258b78f95/source/agora/test/Failures.d#L34-L38 redundant.

### DIFF
--- a/source/geod24/LocalRest.d
+++ b/source/geod24/LocalRest.d
@@ -414,7 +414,8 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
 
        Params:
            Implementation = Type of the implementation to instantiate
-           args = Arguments to `Implementation`'s constructor
+           self = The channel on which to "listen" to receive new "connections"
+           cargs = Arguments to `Implementation`'s constructor
 
     ***************************************************************************/
 
--- a/source/geod24/LocalRest.d
+++ b/source/geod24/LocalRest.d
@@ -317,10 +317,11 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
 
     ***************************************************************************/
 
-    public static RemoteAPI spawn (Impl) (CtorParams!Impl args,
-        Duration timeout = Duration.init)
+    public static RemoteAPI spawn (Impl) (
+        CtorParams!Impl args, Duration timeout = Duration.init,
+        string file = __FILE__, int line = __LINE__)
     {
-        auto childTid = C.spawn(&spawned!(Impl), args);
+        auto childTid = C.spawn(&spawned!(Impl), file, line, args);
         return new RemoteAPI(childTid, timeout);
     }
 
@@ -403,11 +404,14 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
        Params:
            Implementation = Type of the implementation to instantiate
            self = The channel on which to "listen" to receive new "connections"
+           file = Path to the file that spawned this node
+           line = Line number in the `file` that spawned this node
            cargs = Arguments to `Implementation`'s constructor
 
     ***************************************************************************/
 
-    private static void spawned (Implementation) (C.Tid self, CtorParams!Implementation cargs)
+    private static void spawned (Implementation) (
+        C.Tid self, string file, int line, CtorParams!Implementation cargs)
         nothrow
     {
         import std.datetime.systime : Clock, SysTime;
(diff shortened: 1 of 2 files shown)


AndrejMitrovic:
>https://github.com/bpfkorea/agora/blob/766e3239f26f161c9475f8b7355f792258b78f95/source/agora/test/Failures.d#L34-L38 redundant.

That's great. Also I plan to remove that file, since it's testing the quorum configuration, which will be removed.

source/geod24/LocalRest.d
```diff
@@ -502,9 +505,20 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
...
+            if (t is exc)
+                return;
+
+            import core.stdc.stdio, std.stdio;
+            printf("#### INTERNAL ERROR: %.*s\n", cast(int) t.msg.length, t.msg.ptr);
+            printf("Please file a bug at https://github.com/Geod24/localrest/\n");
+
+            try writeln("Full error: ", t);
+            catch (Exception e) { /* Nothing more we can do at this point */ }
+        }
```
AndrejMitrovic, line 528: > Instead of silently loosing a thread, we catch `Throwable` and at least print a message before re-throwing

But the exception is not re-thrown here. Would it makes sense to add an `assert(0)` after the printouts?

source/geod24/LocalRest.d
```diff
@@ -402,11 +403,15 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
 
        Params:
            Implementation = Type of the implementation to instantiate
+           self = The channel on which to "liston" to receive new "connections"
```
```suggestion
           self = The channel on which to "listen" to receive new "connections"
```

source/geod24/LocalRest.d
```diff
@@ -402,11 +403,15 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
...
+           file = Path to the file that spawned this node
+           line = Line number in the `file` that spawned this node
            args = Arguments to `Implementation`'s constructor
```
AndrejMitrovic: I've noticed there's a typo here from before. `args` => `cargs`

source/geod24/LocalRest.d
```diff
@@ -740,7 +740,12 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
 
                         auto command = Command(C.thisTid(), scheduler.getNextResponseId(), ovrld.mangleof,
                                                SerializedData(serialized));
-                        C.send(this.childTid, command);
+
+                        // If the node already shut down, its MessageBox will be
+                        // closed. Detect it and notify the user.
+                        // Note that it might be expected that the remote died.
+                        if (!C.trySend(this.childTid, command))
+                            throw new Exception("Connection with peer closed");
```
AndrejMitrovic, line 748: So it seems like in our case when a node receives its `shutdown` call, we should stop all the timers (the fibers) that are currently running. Like the periodic `getBlocksFrom`.

AndrejMitrovic:
LGTM, just found a typo. Do these changes / features need to be documented in the readme?

source/geod24/LocalRest.d
Geod24: Well re-throwing doesn't do anything. The thread just disappear, without any output.
The only case where an `assert(0)` would make sense, IMO, would be:
- When compiled in `-release`
- When compiled with `-checkaction=halt`

Since localrest is a testing framework, the release part doesn't make much sense. For the second case, changing `checkaction`, well we won't get here to begin with, because we don't throw Exceptions, we only arrive here if there's an `AssertError` (or maybe other errors, e.g. `RangeError`).

source/geod24/LocalRest.d
Geod24: I was originally thinking so, but... Do we really need to ? Those methods should handle client disconnection gracefully anyway.

Geod24:
Added a commit with documentation fixes (it's the second one in the list, no matter what Github says).

Geod24:
Codecov isn't happy because some `assert(0)` are uncovered. Well too bad.

AndrejMitrovic:
Should we merge it regardless then?

Geod24:
I think so yes. I'll change the configuration not to show red on Codecov `X`.

PR #67, opened 2020-05-19
Call an optional cleanup routine before a node is destroyed

AndrejMitrovic:
I've realized what was causing the segfaults in our Agora test-suite. It's a race condition between these calls:

```
// e.g. calls SQL shutdown methods which cannot be called in the dtor
node.shutdown();  

// race condition *here*

// destroys Node, shuts down event loop
node.ctrl.shutdown();
```

The logging confirmed it.

I've implemented a simple fix in this PR. A nice plus is that `shutdown` no longer has to be an `API` endpoint, and we can remove it from our interfaces in Agora.

### DIFF
--- a/source/geod24/LocalRest.d
+++ b/source/geod24/LocalRest.d
@@ -40,6 +40,21 @@
     methods, for some defined time or until unblocked.
     See `sleep`, `filter`, and `clearFilter` for more details.
 
+    Shutdown:
+    The control interface has a shutdown method that can be used to terminate
+    a node gracefully. When the shutdown request is handled by the node,
+    the event loop will exit and the thread will terminate. While the destructor
+    of the node will be called, it might not usable for some actions, for example
+    because D destructors may not allocate GC memory, or one may want
+    to perform some test-specific operations, such a logging some data in case of failure.
+    Therefore, you may provide a shutdown routine in the call to `shutdown`.
+    It must accept a single argument of the interface type, and will be called
+    with the implementation object just before the node is destroyed.
+    If this routine throws, LocalRest will log the error in the console and
+    proceed with destroying the stack-allocated node.
+    Note that control requests are asynchronous, hence requests from the node
+    might be processed / send by the node until the request is actually processed.
+
     Event_Loop:
     Server process usually needs to perform some action in an asynchronous way.
     Additionally, some actions needs to be completed at a semi-regular interval,
@@ -116,8 +131,10 @@ private struct TimeCommand
 }
 
 /// Ask the node to shut down
-private struct ShutdownCommand
+private struct ShutdownCommand (API)
 {
+    /// Any callback to call before the Node's destructor is called
+    void function (API) callback;
 }
 
 /// Filter out requests before they reach a node
@@ -481,7 +498,13 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
                 while (1)
                 {
                     C.receiveTimeout(self, 10.msecs,
-                        (ShutdownCommand e) { throw exc; },
+                        (ShutdownCommand!API e)
+                        {
+                            if (e.callback !is null)
+                                e.callback(node);
+
+                            throw exc;
+                        },
                         (TimeCommand s)      {
                             control.sleep_until = Clock.currTime + s.dur;
                             control.drop = s.drop;
@@ -594,11 +617,17 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
 
             Send an async message to the thread to immediately shut down.
 
+            Params:
+                callback = if not null, the callback to call in the Node's
+                           thread before the Node is destroyed. Can be used
+                           for cleanup / logging routines.
+
         ***********************************************************************/
 
-        public void shutdown () @trusted
+        public void shutdown (void function (API) callback = null)
+            @trusted
         {
-            C.send(this.childTid, ShutdownCommand());
+            C.send(this.childTid, ShutdownCommand!API(callback));
         }
 
         /***********************************************************************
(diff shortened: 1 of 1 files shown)


AndrejMitrovic:
tl;dr: We will replace `node.shutdown && node.ctrl.shutdown` with just `node.ctr.shutdown` in Agora, and also remove `shutdown` from its API.

AndrejMitrovic:
Fixed up the implementation according to tips by @Geod24.

Note that I couldn't use the full type from the `spawn` call because this type is only defined in the instantiation of `spawned!(Implementation)` and it's not available in the outer `RemoteAPI!(API)`. So client code will have to downcast if they wish to call methods not in the API.

source/geod24/LocalRest.d
```diff
@@ -393,6 +398,13 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
         }
     }
 
+    // Ask the node to shut down. Kept here as we need access to `API` symbol.
+    private static struct ShutdownCommand
+    {
+        /// Any callback to call before the Node's destructor is called
+        void function (API) callback;
```
AndrejMitrovic: I deliberately omitted `nothrow` because exceptions are already caught and logged in LocalRest. So it's nicer UX.

source/geod24/LocalRest.d
```diff
@@ -40,6 +40,16 @@
     methods, for some defined time or until unblocked.
     See `sleep`, `filter`, and `clearFilter` for more details.
 
+    Shutdown:
+    The control interface has a shutdown method. It may be called when one
+    wants to shut down the node and call its destructor. Additionally some
+    classes may wish to do resource deallocation separately from their
+    destructors, in particular when those methods themselves allocate
+    (D destructors may not allocate GC memory). Therefore, you may provide
```
Geod24: That is actually not a concern for localrest. The node is allocated using `scope`, which means it's stack allocated, which means the destructor runs deterministically and can allocate.

source/geod24/LocalRest.d
```diff
@@ -40,6 +40,16 @@
...
+    a shutdown routine in the call to `shutdown`, it will be called before
+    the Node is destroyed. This `shutdown` method should not schedule any
```
Geod24: Why The Capitalization Of Node ?

source/geod24/LocalRest.d
```diff
@@ -115,11 +125,6 @@ private struct TimeCommand
     bool drop = false;
 }
 
-/// Ask the node to shut down
-private struct ShutdownCommand
```
Geod24, line 119: Why did you move it ? Just template it.

source/geod24/LocalRest.d
```diff
@@ -823,6 +847,46 @@ unittest
...
+
+    static interface API
+    {
+        @safe:
+        public @property ulong pubkey ();
+    }
+
+    static class MockAPI : API
+    {
+        @safe public override @property ulong pubkey ()
```
```suggestion
        public override @property ulong pubkey () @safe
```

source/geod24/LocalRest.d
```diff
@@ -823,6 +847,46 @@ unittest
...
+        auto mock = cast(MockAPI)node;
+        assert(mock !is null);
+        mock.shutdown();
+        onDestroy_called = true;
+    }
+
+    scope test = RemoteAPI!API.spawn!MockAPI();
+    assert(test.pubkey() == 42);
+    test.ctrl.shutdown(&onDestroy);
+    core.thread.Thread.sleep(1.seconds);  // ctr.shutdown call is asynchronous
```
```suggestion
    while (!core.atomic.atomicLoad(dtor_called)) core.atomic.pause();
```

The problem is that this test doesn't fail, but at least it won't spuriously fail.

source/geod24/LocalRest.d
AndrejMitrovic: I thought if I templatized it you would complain about there being too many templates. :D

Geod24: Well it's a template anyway 🤷

source/geod24/LocalRest.d
AndrejMitrovic: Ah well it was just an example, I could remove it. Anyway how is a destructor supposed to know if the object it's cleaning is allocated on the heap vs the stack?

https://github.com/biozic/d2sqlite3/blob/0280ef1769f0e0907980a241b9ac68e444a1fbd5/source/d2sqlite3/database.d#L89

source/geod24/LocalRest.d
Geod24: Good point

Geod24: Maybe document what happens if one `throw` then ?

source/geod24/LocalRest.d
Geod24: Looking at this method, I thought: "Please don't implement this by catching `InvalidMemoryOperationError`... PLEASE".
[Welp](https://github.com/biozic/d2sqlite3/blob/08e77f5f2da3ead57d62c80226bd85e41c27fbd7/source/d2sqlite3/internal/memory.d#L70)

source/geod24/LocalRest.d
AndrejMitrovic: > core.atomic.pause

hello there

Cool. Didn't know about this.

AndrejMitrovic: We should fix other places where we have sleep calls like this.

AndrejMitrovic: Is `core.atomic.pause` a new feature? I'm getting a build failure on ldc 14 and DMD 2.085.1

AndrejMitrovic: Oh those are quite old compilers too. But I assume you still want to support them.

source/geod24/LocalRest.d
```diff
@@ -40,6 +40,14 @@
     methods, for some defined time or until unblocked.
     See `sleep`, `filter`, and `clearFilter` for more details.
 
+    Shutdown:
+    The control interface has a shutdown method. You may provide a
+    shutdown routine in the call to `shutdown`, it will be called before
+    the node is destroyed. This `shutdown` method should not schedule any
+    tasks / fibers within the node, otherwise it may introduce race conditions.
+    If this routine throws, LocalRest will log the error in the console and
+    proceed with destroying the stack-allocated node.
```
Geod24: I gave a shot at rewriting it:
```suggestion
    Shutdown:
    The control interface has a shutdown method that can be used to terminate
    a node gracefully. When the shutdown request is handled by the node,
    the event loop will exit and the thread will terminate. While the destructor
    of the node will be called, it might not usable for some actions, for example
    because D destructors may not allocate GC memory, or one may want
    to perform some test-specific operations, such a logging some data in case of failure.
    Therefore, you may provide a shutdown routine in the call to `shutdown`.
    It must accept a single argument of the interface type, and will be called
    with the implementation object just before the node is destroyed.
    If this routine throws, LocalRest will log the error in the console and
    proceed with destroying the stack-allocated node.
    Note that control requests are asynchronous, hence requests from the node
    might be processed / send by the node until the request is actually processed.
```

I also think we should add a mention that all control interface methods are asynchronous (in the previous paragraph). I also wanted to focus more on the shutdown method itself than the optional argument, while keeping the limitations that brought us here and you previously documented.

Geod24:
You might need to update the line breaks tho :P

source/geod24/LocalRest.d
Geod24: Yeah a `sleep` in a loop might just be simpler

Geod24: My main concern was the lack of a retry mechanism

AndrejMitrovic:
> You might need to update the line breaks tho :P

I think it looks okay.

Geod24 (approved):

PR #69, opened 2020-05-25
add setTimer function in LocalRest

TrustHenry:
Add a `setTimer` to call the delegate function periodically.
It works similar to the `setTimer` of vibe.d

### DIFF
--- a/source/geod24/LocalRest.d
+++ b/source/geod24/LocalRest.d
@@ -282,6 +282,37 @@ public void sleep (Duration timeout)
     cond.wait(timeout);
 }
 
+/*******************************************************************************
+
+    Run an asynchronous task after a given time.
+
+    The task will first run after the given `timeout`, and
+    can either repeat or run only once (the default).
+    Works similarly to Vibe.d's `setTimer`.
+
+    Params:
+        timeout = Determines the minimum amount of time that elapses before
+            the timer fires.
+        dg = If non-null, this delegate will be called when the timer fires
+        periodic = Speficies if the timer fires repeatedly or only once
+
+*******************************************************************************/
+
+public void setTimer (Duration timeout, void delegate() dg,
+    bool periodic = false)
+{
+    assert(scheduler !is null, "Cannot call this delegate from the main thread");
+    assert(dg !is null, "Cannot call this delegate if null");
+    scheduler.schedule(
+    ()
+    {
+        do
+        {
+            sleep(timeout);
+            dg();
+        } while (periodic);
+    });
+}
 
 /*******************************************************************************
 
@@ -1842,4 +1873,47 @@ unittest
 
     node2.ctrl.shutdown();
     thread_joinAll();
+}
+
+/// Test setTimer
+unittest
+{
+    static import core.thread;
+    import core.time;
+
+    static interface API
+    {
+        public void startTimer ();
+        public ulong getCounter ();
+    }
+
+    static class Node : API
+    {
+        public override void startTimer ()
+        {
+            setTimer(100.msecs, &callback, /* periodic: */ true);
+        }
+
+        public void callback () @safe
+        {
+            this.counter++;
+        }
+
+        public override ulong getCounter ()
+        {
+            scope (exit) this.counter = 0;
+            return this.counter;
+        }
+
+        private ulong counter;
+    }
+
+    auto node = RemoteAPI!API.spawn!Node();
+    assert(node.getCounter() == 0);
+    node.startTimer();
+    core.thread.Thread.sleep(1.seconds);
+    // The expected count is 9
+    // Check means the timer repeated
+    assert(node.getCounter() >= 2);
+    node.ctrl.shutdown();
 }
--- a/source/geod24/concurrency.d
+++ b/source/geod24/concurrency.d
@@ -845,6 +845,24 @@ class FiberScheduler
         FiberScheduler.yield();
     }
 
+    /**************************************************************************
+
+        Schedule a task to be run next time the scheduler yields
+
+        Behave similarly to `spawn`, but instead of running the task
+        immediately, it simply adds it to the queue and continue executing
+        the current task.
+
+        Params:
+            op = Operation to run
+
+    **************************************************************************/
+
+    void schedule(void delegate() op) nothrow
+    {
+        this.create(op);
+    }
+
     /**
      * If the caller is a scheduled Fiber, this yields execution to another
      * scheduled Fiber.


source/geod24/LocalRest.d
```diff
@@ -1843,3 +1860,44 @@ unittest
...
+        }
+
+        public void callback () @safe
+        {
+            this.counter++;
+        }
+
+        public override ulong getCounter ()
+        {
+            scope (exit) this.counter = 0;
```
AndrejMitrovic, line 1904:
```suggestion
```

It seems unnecessary.

source/geod24/LocalRest.d
```diff
@@ -1843,3 +1860,44 @@ unittest
...
+        }
+
+        private ulong counter;
+    }
+
+    auto node = RemoteAPI!API.spawn!Node();
+    assert(node.getCounter() == 0);
+    node.startTimer();
+    core.thread.Thread.sleep(1.seconds);
+    assert(node.getCounter() >= 8);
```
```suggestion
    assert(node.getCounter() >= 2);
```

2 should be enough (it means the timer repeated). 8 is too close to the possible upper limit of 10 (10 calls every 100 msecs), and if the machine is too slow then it might have called the timer fewer times than 8.

source/geod24/LocalRest.d
```diff
@@ -282,6 +282,23 @@ public void sleep (Duration timeout)
     cond.wait(timeout);
 }
 
+/// Ditto
+public void setTimer (Duration timeout, void delegate() dg,
```
AndrejMitrovic, line 301: This function should be documented somewhere. For example, in the module documentation.

source/geod24/LocalRest.d
```diff
@@ -1843,3 +1860,44 @@ unittest
     node2.ctrl.shutdown();
     thread_joinAll();
 }
+
+///
```
```suggestion
/// Test setTimer
```

source/geod24/concurrency.d
```diff
@@ -845,6 +845,23 @@ class FiberScheduler
...
+        Schedule a task to be run next time the scheduler yields
+
+        Behave similarly to `spawn`, but instead of running the task
+        immediately, it simply adds it to the queue and continue executing
+        the current task.
+
+        Params:
+            op = Operation to run
+
+    **************************************************************************/
```
AndrejMitrovic, line 859:
```suggestion
    **************************************************************************/

```

source/geod24/LocalRest.d
```diff
@@ -282,6 +282,23 @@ public void sleep (Duration timeout)
     cond.wait(timeout);
 }
 
+/// Ditto
+public void setTimer (Duration timeout, void delegate() dg,
+    bool periodic = true)
+{
+    assert(scheduler !is null, "Cannot call this function from the main thread");
+    assert(dg !is null, "Cannot call this delegate function if null");
```
```suggestion
    assert(dg !is null, "Cannot call this delegate if null");
```

source/geod24/LocalRest.d
```diff
@@ -282,6 +282,23 @@ public void sleep (Duration timeout)
     cond.wait(timeout);
 }
 
+/// Ditto
+public void setTimer (Duration timeout, void delegate() dg,
+    bool periodic = true)
+{
+    assert(scheduler !is null, "Cannot call this function from the main thread");
```
```suggestion
    assert(scheduler !is null, "Cannot call this delegate from the main thread");
```

AndrejMitrovic (changes requested):

source/geod24/LocalRest.d
AndrejMitrovic: Ah you reset it on purpose. Please disregard my comment.

source/geod24/LocalRest.d
TrustHenry: Apply the suggestion.

TrustHenry:
@AndrejMitrovic All suggestions have been applied. Thanks to your reviews

AndrejMitrovic (approved):

source/geod24/LocalRest.d
```diff
@@ -259,8 +259,8 @@ private bool is_main_thread;
     usually provided by `vibe.core.core : runTask`.
 
     In order for them to properly work, we need to integrate them to our event
-    loop by providing the ability to spawn a task, and wait on some condition,
-    optionally with a timeout.
+    loop by providing the ability to spawn a task, and wait and timer
+    on some condition, optionally with a timeout.
```
Geod24: That sentence does not compute

source/geod24/LocalRest.d
```diff
@@ -282,6 +282,23 @@ public void sleep (Duration timeout)
...
+    bool periodic = true)
+{
+    assert(scheduler !is null, "Cannot call this delegate from the main thread");
+    assert(dg !is null, "Cannot call this delegate if null");
+    scheduler.schedule(
+    ()
+    {
+        do
+        {
+            // periodic task
```
Geod24: Is there a use to that comment ?

Geod24 (commented):
Only documentation issue, oLGTM

source/geod24/LocalRest.d
TrustHenry: Deleted the existing phrase and then Function document added

source/geod24/LocalRest.d
TrustHenry: No, I deleted it.

source/geod24/LocalRest.d
```diff
@@ -282,6 +282,35 @@ public void sleep (Duration timeout)
...
+
+    Run an asynchronous task after the set time in LocalRest's
+    can set it to repeat or run it only once.
+    Works similar to the setTimer of vibe.d.
+
+    Params:
+        timeout = Determines the minimum amount of time that elapses before
+            the timer fires.
+        dg = If non-null, this delegate will be called when the timer fires
+        periodic = Speficies if the timer fires repeatedly or only once
```
```suggestion
    Run an asynchronous task after a given time.

    The task will first run after the given `timeout`, and
    can either repeat or run only once (the default).
    Works similarly to Vibe.d's `setTimer`.

    Params:
        timeout = Determines the minimum amount of time that elapses before
            the timer fires.
        dg = If non-null, this delegate will be called when the timer fires
        periodic = Speficies if the timer fires repeatedly or only once
```

source/geod24/LocalRest.d
```diff
@@ -282,6 +282,35 @@ public void sleep (Duration timeout)
...
+
+*******************************************************************************/
+
+public void setTimer (Duration timeout, void delegate() dg,
+    bool periodic = true)
```
Geod24: [setTimer](https://vibed.org/api/vibe.core.core/setTimer) is non-periodic by default, and I think that makes more sense.

TrustHenry: It's with periodic focus, but I'll apply the suggestion.

Geod24 (approved):

AndrejMitrovic (approved):

PR #71, opened 2020-05-26
Add ability to restart a node

Geod24:
CC @linked0

### DIFF
--- a/source/geod24/LocalRest.d
+++ b/source/geod24/LocalRest.d
@@ -37,8 +37,8 @@
     a few extra functionalities are offered under a control interface,
     accessible under the `ctrl` namespace in the instance.
     The control interface allows to make the node unresponsive to one or all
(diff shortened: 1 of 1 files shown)


Geod24:
When pair programming with @linked0 yesterday we realized that actually restarting a node was not quite possible yet, but it was obvious it should be, and it should be easy to implement.

We had multiple options, including the option to replace the client with a "lightweight" client which can have its `Tid` replaced, but implementing it in LocalRest seemed like the best option (also because it solves the issue of the client code having to know about the underlying type).

In the future we might want to provide a mean to replace a node (e.g. replace a well-behaving node with a byzantine one) but we will tackle that in due time.

linked0 (approved):
LGTM

linked0:
Can this be merged if there is no big problem? I have a plan to use the `restart` today.

PR #72, opened 2020-05-27
Implement withTimeout() for cloning RemoteAPI with a custom timeout

AndrejMitrovic:
Thanks to @Geod24 for the idea.

### DIFF
--- a/source/geod24/LocalRest.d
+++ b/source/geod24/LocalRest.d
@@ -39,6 +39,9 @@
     The control interface allows to make the node unresponsive to one or all
     methods, for some defined time or until unblocked, as well as trigger
     shutdowns or restart. See the methods for more details.
+    The `withTimeout` control method can be used to spawn a scoped copy
+    of the RemoteAPI with a custom configured timeout. The user-provided
+    delegate will be called with this scoped copy that uses the new timeout.
 
     Shutdown:
     The control interface has a shutdown method that can be used to terminate
@@ -820,6 +823,30 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
         {
             C.send(this.childTid, FilterAPI(""));
         }
+
+        /***********************************************************************
+
+            Call the provided delegate with a custom timeout
+
+            This allow to perform requests on a client with a different timeout,
+            usually to allow some requests (e.g. initialization calls) to have longer
+            timeout, or no timeout at all, or to put a timeout on an otherwise
+            timeout-less client (e.g. when calling the actual test which could fail).
+
+            To disable timeout, pass the special value `Duration.zero`
+            (or `0.seconds`, `0.msecs`, etc...).
+
+            Params:
+                timeout = the new timeout to use
+                dg = the delegate to call with the new scoped RemoteAPI copy
+
+        ***********************************************************************/
+
+        public void withTimeout (Dg) (Duration timeout, scope Dg dg) @trusted
+        {
+            scope api = new RemoteAPI(this.childTid, timeout);
+            dg(api);
+        }
     }
 
     /***************************************************************************
@@ -1500,6 +1527,28 @@ unittest
     // node with no timeout
     auto node = RemoteAPI!API.spawn!Node();
     assert(node.sleepFor(80) == 42);  // no timeout
+
+    // custom timeout
+    bool called;
+    node.ctrl.withTimeout(100.msecs,
+        (scope RemoteAPI!API api) {
+            assertThrown!Exception(api.sleepFor(2000));
+            called = true;
+        });
+    assert(called);
+
+    called = false;
+    struct S
+    {
+        void opCall (scope RemoteAPI!API api)
+        {
+            assertThrown!Exception(api.sleepFor(2000));
+            called = true;
+        }
+    }
+    S s;
+    node.ctrl.withTimeout(100.msecs, s);
+    assert(called);
 
     // node with a configured timeout
     auto to_node = RemoteAPI!API.spawn!Node(500.msecs);


AndrejMitrovic:
I also tried the `opDispatch` way. And while it works, it will probably create too much template spam.

I did have a bug in my test-code at first. I called the wrong API instead of the one provided to the delegate. So one needs to be a bit careful when using this..

source/geod24/LocalRest.d
```diff
@@ -820,6 +823,24 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
...
+            with a custom timeout.
+
+            Params:
+                timeout = the new timeout to use
+                dg = the delegate to call with the new scoped RemoteAPI copy
+
+        ***********************************************************************/
+
+        public void withTimeout (Duration timeout,
+            void delegate (scope RemoteAPI) dg) @trusted
```
```suggestion
            scope void delegate (scope RemoteAPI) dg) @trusted
```

source/geod24/LocalRest.d
```diff
@@ -1501,6 +1522,12 @@ unittest
     auto node = RemoteAPI!API.spawn!Node();
     assert(node.sleepFor(80) == 42);  // no timeout
 
+    // custom timeout
+    node.ctrl.withTimeout(500.msecs,
+        (scope api) {
```
AndrejMitrovic: If I make the delegate templated in `withDuration`, then I must specify the correct type here. 

`(scope RemoteAPI!API api)` rather than just `(scope api)`

AndrejMitrovic: On the other hand, making it templated allows passing any callable type.

AndrejMitrovic:
Ok I made it templated. I'm not sure if you want it more or less constrained.

source/geod24/LocalRest.d
```diff
@@ -820,6 +823,24 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
...
+            Call the provided delegate with a scoped copy of this RemoteAPI
+            with a custom timeout.
+
+            Params:
+                timeout = the new timeout to use
+                dg = the delegate to call with the new scoped RemoteAPI copy
+
+        ***********************************************************************/
+
+        public void withTimeout (Dg) (Duration timeout, Dg dg) @trusted
```
```suggestion
        public void withTimeout (Dg) (Duration timeout, scope Dg dg)
```

AndrejMitrovic:
Hmm this template constraint is not always passing, when used in Agora. I'll have to fix this.

source/geod24/LocalRest.d
```diff
@@ -101,7 +104,7 @@ module geod24.LocalRest;
 static import C = geod24.concurrency;
 import geod24.Serialization;
 import std.meta : AliasSeq;
-import std.traits : Parameters, ReturnType;
+import std.traits : isDelegate, Parameters, ReturnType;
```
Geod24: Leftover ? You probably wanted `isCallable` right ?

source/geod24/LocalRest.d
Geod24: By the way the signature I had originally in mind was `withTimeout(alias Callable)(Duration timeout)` but this works too

AndrejMitrovic:
> Hmm this template constraint is not always passing, when used in Agora. I'll have to fix this.

It's because it was missing the `S`, serializer. But I can just write it as `RemoteAPI` instead of manually instantiating it, because that symbol refers to the instantiation already. Fixed.

source/geod24/LocalRest.d
```diff
@@ -820,6 +823,24 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
         {
             C.send(this.childTid, FilterAPI(""));
         }
+
+        /***********************************************************************
+
+            Call the provided delegate with a scoped copy of this RemoteAPI
+            with a custom timeout.
+
```
```suggestion
            Call the provided delegate with a custom timeout

            This allow to perform requests on a client with a different timeout,
            usually to allow some requests (e.g. initialization calls) to have longer
            timeout, or no timeout at all, or to put a timeout on an otherwise
            timeout-less client (e.g. when calling the actual test which could fail).
            
            To disable timeout, pass the special value `Duration.zero`
            (or `0.seconds`, `0.msecs`, etc...).
```

source/geod24/LocalRest.d
```diff
@@ -820,6 +823,24 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
...
+            Params:
+                timeout = the new timeout to use
+                dg = the delegate to call with the new scoped RemoteAPI copy
+
+        ***********************************************************************/
+
+        public void withTimeout (Dg) (Duration timeout, Dg dg) @trusted
+            if (is(typeof({ dg(RemoteAPI.init); })))
```
Geod24: In general, when there is no overload involved, I prefer to use `static assert` in the body of the function to provide a more friendly error message. Template constraints should only really be used to drive overload / specialization resolution.

AndrejMitrovic: It's because template constraints are terrible. :>

AndrejMitrovic: Diagnostics I mean.

source/geod24/LocalRest.d
AndrejMitrovic: Haha, you got me.

source/geod24/LocalRest.d
AndrejMitrovic: It's unfortunate ldc2 errors with the "need dual context support", otherwise yes it would look really nice that way.

Geod24 (approved):

PR #75, opened 2020-05-27
Implement scheduleTask()

AndrejMitrovic:
It could perhaps use some documentation?

Let me know what you think @Geod24 (as always :P).

### DIFF
--- a/source/geod24/LocalRest.d
+++ b/source/geod24/LocalRest.d
@@ -938,6 +938,44 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
         }
 }
 
+/// test runTask vs scheduleTask behavior
+unittest
+{
+    import std.stdio;
+
+    static interface API
+    {
+        public void start ();
+    }
+
+    static class Node : API
+    {
+        int[] calls;
+
+        public override void start ()
+        {
+            writefln("This: %s\n", cast(void*)this);
+
+            scheduleTask(
+            {
+                writefln("This: %s\n", cast(void*)this);
+                calls ~= 11;
+            });
+        }
+
+        ~this ()
+        {
+            printf("Node.dtor\n");
+        }
+    }
+
+    auto node = RemoteAPI!API.spawn!Node();
+    node.start();
+    node.ctrl.shutdown();
+}
+
+version (none):
+
 /// Simple usage example
 unittest
 {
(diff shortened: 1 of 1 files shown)


Geod24:
What's the `Vibe.d` counterpart ?

AndrejMitrovic:
> What's the Vibe.d counterpart ?

Good question! I'll experiment with it and get back to you on that one.

Btw the segfaults are bizarre. I wonder if the task runs after the object was collected..

AndrejMitrovic:
> What's the Vibe.d counterpart ?

Ah you mean what is the name of that function. Well actually I am still looking for it. That's a very good point though, there's not much point in supporting that if it doesn't exist.

AndrejMitrovic:
Hmm. Maybe it actually makes sense that no equivalent in vibe.d exists. If a fiber starts a new task, it probably wants the "child" fiber to run before the "parent" fiber dies.

So maybe we need to get rid of `FiberScheduler.schedule`.. it could be considered unsafe to use?

Geod24:
>  If a fiber starts a new task, it probably wants the "child" fiber to run before the "parent" fiber dies.

Why so ?

Geod24:
For me there are two separate issues: Starting tasks, and synchronizing them. You have cases where you don't want them synchronized, and times when you want them synchronized.

Example:
1) Receive a request => Start a task (new fiber) => Do some computing and schedule a timer to check XXX in 30 minutes.
In this case you don't want the task that is scheduled (the timer) to die when you parent dies (the request handler returns).

2) You receive a request, and you need to get data from 42 different servers to complete it. You schedule 42 different requests and then wait on all of them.
In this case, you do want the parent to outlive the children. But you want to schedule **all** your I/O before actually doing the context switch / synchronization.

Geod24:
Another note: We could totally have LocalRest be single-threaded.

I originally thought it wasn't possible, but now that we implemented everything more or less correctly, I think it could just work. But test would be too deterministic (and probably slower).
See those few sentences we removed a few months ago which betray my original mistake: https://github.com/Geod24/localrest/commit/756af07d51ed7deadb912d753ff1f2bf58e3e4b5#diff-b5229c55d46fb0a852c49674585a3595L16

AndrejMitrovic:
> But test would be too deterministic (and probably slower).

We could randomize the order the fibers from different nodes run in, I guess? Just thinking out loud..

AndrejMitrovic:
Right, the "parent" fiber dying shouldn't be a problem.

> What's the Vibe.d counterpart ?

Couldn't find anything, but it's very simple to implement:

```D
// in TaskManager
void runTask (void delegate() dg) 
{ 
    vibed.runTask({ vibed.yield(); dg(); });
}
```

> For me there are two separate issues: Starting tasks, and synchronizing them. 

Yep. We don't currently have support for waiting on a specific task like we can in vibe.d. We probably want to add support for this later.

Geod24:
That Vibe.d example is not quite the same though. You are not guaranteed to return to your task if you yield immediately.

AndrejMitrovic:
Hmm that's a good point. 

I'm looking at other options. There is `switchToTask`:

```D
void runTask (void delegate() dg) 
{ 
    auto parent_task = Task.getThis();
    vibed.runTask({ switchToTask(parent_task); dg(); });
}
```

However I'm not sure whether it's legal to call `switchToTask` to a task that yielded (the docs mention it should be `hibernate()'d`), and the docs also don't mention when the child task will be switched back to.

AndrejMitrovic:
Btw, as I mentioned on Slack the implementation of runTask in LocalRest and vibe.d is different. LocalRest appends the task, suspends the calling fiber, but this doesn't guarantee the just added task will be ran next.

Vibe.d immediately starts the task in runTask().

But I think you mentioned you're working on refactoring the FiberScheduler right?

AndrejMitrovic:
Do you think Sönke is receptive of adding new features to vibe.d? I feel like scheduling but not running a task makes sense to be supported.

AndrejMitrovic:
Still debugging, but I think I found some very interesting DMD bug. This segfaults:

```D
unittest
{
    import std.stdio;

    static interface API
    {
        public void start ();
    }

    static class Node : API
    {
        int[] calls;

        public override void start ()
        {
            writefln("This: %s\n", cast(void*)this);

            scheduleTask(
            {
                writefln("This: %s\n", cast(void*)this);
                calls ~= 11;
            });
        }

        ~this ()
        {
            printf("Node.dtor\n");
        }
    }

    auto node = RemoteAPI!API.spawn!Node();
    node.start();
    node.ctrl.shutdown();
}
```

The `this` instance is `null` inside of the scheduled fiber. Only when compiled with DMD.

Geod24:
>

Yep. This is some of the documentation for what I'm working on:

>    The scheduler exposes the following primitives:
>    - schedule: Add a task to the set of tasks to execute;
>    - run: Run a new task and block the current task until completion;
>    - yield(id): Switch execution to the provided `id`
>    - dispatch: Yield from the current task and perform automatic dispatch
>
>    When a scheduler is instantiated, the program is not in a fiber.
>    `run` should be called first. Using `schedule` followed to either
>    `dispatch` or `yield` with one of the scheduled task's `id` is also
>    supported. However calling `yield` with a non-scheduled `id` or
>    `dispatch` without any scheduled task will trigger an `abort`.

AndrejMitrovic:
Then I can just close this, since it's not urgent to have yet. But I'll still try to find the root cause of the segfault..

PR #76, opened 2020-05-28
Fix usage of scoped delegate in the runTask() API

AndrejMitrovic:


Geod24:
WIP?

AndrejMitrovic:
Sorry I accidentally added `version (none):` to the other tests. Fixed. 😇

source/geod24/LocalRest.d
```diff
@@ -931,6 +931,45 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
         }
 }
 
+/// test for scoped delegates
+unittest
```
Geod24: Can you move this towards the end please ? The first example is meant to be / stay a simple usage example. Also I wouldn't make this a documented test

AndrejMitrovic:
Good idea to move it elsewhere. And we should probably move the tests to another module as LocalRest.d is getting quite big.

AndrejMitrovic:
done

PR #78, opened 2020-05-28
Make all tests use thread_joinAll

Geod24:
While not required (tests are independent and so are nodes),
it is more hygienic to do so at the end of the test.
In addition, if a node lives on by mistake, it'll be easier
to track down (because we'll block on the test and get a stack trace).
The final advantage is that it makes things easier to read in debuggers,
as only the threads used for the current tests are visible.

### DIFF
--- a/source/geod24/LocalRest.d
+++ b/source/geod24/LocalRest.d
@@ -959,6 +959,7 @@ unittest
     scope test = RemoteAPI!API.spawn!MockAPI();
     assert(test.pubkey() == 42);
     test.ctrl.shutdown();
+    thread_joinAll();
 }
 
 /// Example where a shutdown() routine must be called on a node before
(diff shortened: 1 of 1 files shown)


source/geod24/LocalRest.d
```diff
@@ -995,8 +996,9 @@ unittest
     scope test = RemoteAPI!API.spawn!MockAPI();
     assert(test.pubkey() == 42);
     test.ctrl.shutdown(&onDestroy);
+    thread_joinAll();
     // ctr.shutdown call is asynchronous
-    while (!dtor_called) Thread.sleep(100.msecs);
+    assert(dtor_called);
```
Geod24, line 1001: Also a small improvement ^_^

AndrejMitrovic (approved):

PR #80, opened 2020-06-04
Small improvements 

AndrejMitrovic:
Styling fix + getting rid of thread.sleep calls where they're unnecessary.

### DIFF
--- a/source/geod24/LocalRest.d
+++ b/source/geod24/LocalRest.d
@@ -549,7 +549,7 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
                             exc.restart = e.restart;
                             throw exc;
                         },
-                        (TimeCommand s)      {
+                        (TimeCommand s) {
                             control.sleep_until = Clock.currTime + s.dur;
                             control.drop = s.drop;
                         },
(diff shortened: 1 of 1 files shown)


source/geod24/LocalRest.d
```diff
@@ -1576,7 +1579,9 @@ unittest
     assert(to_node.sleepFor(40) == 42);
 
     assertThrown!Exception(to_node.sleepFor(2000));
-    Thread.sleep(2.seconds);  // need to wait for sleep() call to finish before calling .shutdown()
+    to_node.ctrl.withTimeout(Duration.zero,  // wait for the node to wake up
```
Geod24: Can we keep the timeout to 2 seconds though ?

AndrejMitrovic: Wouldn't that suffer from timing issues? Maybe 3 would be better?

Geod24: Well, so would the previous version of the code, no ? But 3 works for me too, or even higher. I just prefer to have a timeout, as having an infinite wait when a test doesn't pass it not very developer-friendly.

AndrejMitrovic: Indeed. It's a good catch. I've updated it.

Geod24 (approved):

PR #82, opened 2020-06-12
Add stop function of setTimer

TrustHenry:
Setimer's function starts infinite loop with whether periodic is present or not.
As one of the functional improvements,
we need the ability to stop it under certain conditions

Includes @AndrejMitrovic ideas

### DIFF
--- a/source/geod24/LocalRest.d
+++ b/source/geod24/LocalRest.d
@@ -303,22 +303,59 @@ public void sleep (Duration timeout)
         dg = If non-null, this delegate will be called when the timer fires
         periodic = Speficies if the timer fires repeatedly or only once
 
+    Returns:
+        A `Timer` instance with the ability to control the timer
+
 *******************************************************************************/
 
-public void setTimer (Duration timeout, void delegate() dg,
+public Timer setTimer (Duration timeout, void delegate() dg,
     bool periodic = false)
 {
     assert(scheduler !is null, "Cannot call this delegate from the main thread");
     assert(dg !is null, "Cannot call this delegate if null");
-    scheduler.schedule(
-    ()
+
+    Timer timer = new Timer(timeout, dg, periodic);
+    scheduler.schedule(&timer.run);
+    return timer;
+}
+
+/// Simple timer
+public final class Timer
+{
+    private Duration timeout;
+    private void delegate () dg;
+    // Whether this timer is repeating
+    private bool periodic;
+    // Whether this timer was stopped
+    private bool stopped;
+
+    public this (Duration timeout, void delegate() dg, bool periodic)
+    {
+        this.timeout = timeout;
+        this.dg = dg;
+        this.periodic = periodic;
+        this.stopped = false;
+    }
+
+    // Run a delegate after timeout, and until this.periodic is false
+    private void run ()
     {
         do
         {
             sleep(timeout);
+            if (this.stopped)
+                return;
             dg();
-        } while (periodic);
-    });
+        } while (this.periodic);
+    }
+
+    /// Stop the timer. The next time this timer's fiber wakes up
+    /// it will exit the run() function.
+    public void stop ()
+    {
+        this.stopped = true;
+        this.periodic = false;
+    }
 }
 
 /*******************************************************************************
@@ -1985,7 +2022,7 @@ unittest
     thread_joinAll();
 }
 
-/// Test setTimer
+/// Test Timer
 unittest
 {
     static import core.thread;
@@ -1993,20 +2030,32 @@ unittest
 
     static interface API
     {
-        public void startTimer ();
+        public void startTimer (bool periodic);
+        public void stopTimer ();
         public ulong getCounter ();
+        public void resetCounter ();
     }
 
     static class Node : API
     {
-        public override void startTimer ()
+        private ulong counter;
+        private Timer timer;
+
+        public override void startTimer (bool periodic)
         {
-            setTimer(100.msecs, &callback, /* periodic: */ true);
+            this.timer = setTimer(100.msecs, &callback, periodic);
         }
 
-        public void callback () @safe
+        public override void stopTimer ()
+        {
+            this.timer.stop();
+        }
+
+        public void callback ()
         {
             this.counter++;
+            if (this.counter == 3)
+                this.timer.stop();
         }
 
         public override ulong getCounter ()
@@ -2015,16 +2064,24 @@ unittest
             return this.counter;
         }
 
-        private ulong counter;
+        public override void resetCounter ()
+        {
+            this.counter = 0;
+        }
     }
 
     auto node = RemoteAPI!API.spawn!Node();
     assert(node.getCounter() == 0);
-    node.startTimer();
+    node.startTimer(true);
     core.thread.Thread.sleep(1.seconds);
-    // The expected count is 9
-    // Check means the timer repeated
-    assert(node.getCounter() >= 2);
+    // The expected count is 3
+    // Check means the timer repeated and the timer stoped
+    assert(node.getCounter() == 3);
+    node.resetCounter();
+    node.startTimer(false);
+    node.stopTimer();
+    core.thread.Thread.sleep(500.msecs);
+    assert(node.getCounter() == 0);
     node.ctrl.shutdown();
 }
 


AndrejMitrovic:
It's loop, not roop.

source/geod24/LocalRest.d
```diff
@@ -303,22 +303,52 @@ public void sleep (Duration timeout)
...
         do
         {
             sleep(timeout);
             dg();
-        } while (periodic);
-    });
+        } while (*this.periodic);
+    }
+
+    // Delegate call stop
```
AndrejMitrovic: This needs to be a ddoc comment with `///`.

```suggestion
    /// Stop the timer. The next time this timer's fiber wakes up
    /// it will exit the run() function.
```

source/geod24/LocalRest.d
```diff
@@ -303,22 +303,52 @@ public void sleep (Duration timeout)
...
+
+    public this (Duration timeout, void delegate() dg, bool periodic)
+    {
+        this.timeout = timeout;
+        this.dg = dg;
+        this.periodic = new bool;
+        *this.periodic = periodic;
+    }
+
+    // Run an delegate after timeout
```
```suggestion
    // Run a delegate after timeout, and until this.periodic is false
```

source/geod24/LocalRest.d
```diff
@@ -303,22 +303,52 @@ public void sleep (Duration timeout)
         dg = If non-null, this delegate will be called when the timer fires
         periodic = Speficies if the timer fires repeatedly or only once
 
+    Return:
+        Timer Reference with stop function
```
```suggestion
    Returns:
        a Timer instance with the ability to stop the timer
```

AndrejMitrovic: Notice it's `Returns`, not `Return` by the way.

source/geod24/LocalRest.d
```diff
@@ -303,22 +303,52 @@ public void sleep (Duration timeout)
...
+    scheduler.schedule(() { timer.run(); });
+    return timer;
+}
+
+/// Simple timer
+public struct Timer
+{
+    private Duration timeout;
+    private void delegate () dg;
+    private bool *periodic;
```
```suggestion
    private bool* periodic;
```

AndrejMitrovic: Also I would add a small non-ddoc comment here, to make it obvious:

```D
// Using a pointer so Timer may be copied and stop() will still work
private bool* periodic;
```

AndrejMitrovic (changes requested):
Looks good, just found a few small issues.

TrustHenry:
Thank you. I applied @AndrejMitrovic suggestions.

linked0 (approved):
LGTM

AndrejMitrovic (approved):
LGTM.

Geod24:
> It's loop, not roop.

Sorry

source/geod24/LocalRest.d
```diff
@@ -303,22 +303,54 @@ public void sleep (Duration timeout)
         dg = If non-null, this delegate will be called when the timer fires
         periodic = Speficies if the timer fires repeatedly or only once
 
+    Returns:
+        a Timer instance with the ability to stop the timer
```
Geod24: I think it's important to separate the reason leading to the immediate change from the whole system architecture.
What do  mean by this ?
1) This PR changes `setTimer`.
2) Before it didn't return anything, now it returns `void`.
3) The reason we are doing this is because we want the ability to stop a `Timer`.

However, the third point is not the whole story. If we *only* wanted the ability to stop a timer, we could return a `delegate`. But what happened instead is that we got new information, and changed the system accordingly.

To summarize this long explanation: Don't think about what change you are doing now, but how the system look as a whole after your change. Document your change in the commit, document the system in the code.

In this case:
```suggestion
        A `Timer` instance with the ability to control the timer
```

source/geod24/LocalRest.d
```diff
@@ -303,22 +303,54 @@ public void sleep (Duration timeout)
...
-public void setTimer (Duration timeout, void delegate() dg,
+public Timer setTimer (Duration timeout, void delegate() dg,
     bool periodic = false)
 {
     assert(scheduler !is null, "Cannot call this delegate from the main thread");
     assert(dg !is null, "Cannot call this delegate if null");
-    scheduler.schedule(
-    ()
+
+    Timer timer = Timer(timeout, dg, periodic);
```
Geod24: I think we discussed storing those in the scheduler, didn't we ?

source/geod24/LocalRest.d
```diff
@@ -303,22 +303,54 @@ public void sleep (Duration timeout)
...
+    private void delegate () dg;
+    // Using a pointer so Timer may be copied and stop() will still work
+    private bool* periodic;
+
+    public this (Duration timeout, void delegate() dg, bool periodic)
+    {
+        this.timeout = timeout;
+        this.dg = dg;
+        this.periodic = new bool;
+        *this.periodic = periodic;
```
```suggestion
        this.periodic = new bool(periodic);
```

Should work.

source/geod24/LocalRest.d
```diff
@@ -303,22 +303,54 @@ public void sleep (Duration timeout)
...
+    return timer;
+}
+
+/// Simple timer
+public struct Timer
+{
+    private Duration timeout;
+    private void delegate () dg;
+    // Using a pointer so Timer may be copied and stop() will still work
+    private bool* periodic;
```
Geod24: Since it has an identity, I'm wondering why we don't use classes for this ?

source/geod24/LocalRest.d
TrustHenry: I didn't understand the benefits of storing Timer in the scheduler.

Are you saying that It assigns one Timer to one scheduler?
Or does this control Timer in the schedule?

source/geod24/LocalRest.d
TrustHenry: I understand what you're saying.
I thought the function could expand. But I couldn't document it.

source/geod24/LocalRest.d
TrustHenry: I'd better change it to `class`
because it should have each `periodic`'s pointer

AndrejMitrovic: What he means is this:

```D
final class Timer
{
    bool periodic;

    ....
}
```

With this `periodic` does not need to be a pointer.

source/geod24/LocalRest.d
AndrejMitrovic: I think that was the alternative implementation idea.

source/geod24/LocalRest.d
TrustHenry: > With this periodic does not need to be a pointer.

Oh, thank you. the right answer 'final class'

I was testing in `agora`, and I experienced a change of reference address...

AndrejMitrovic: It shouldn't really matter if it's final or not, it would work either way. But with `final` you cannot inherit from this class.

By the way, what code failed for you?

AndrejMitrovic (changes requested):
This will need to be changed after all. I'll suggest some changes.

source/geod24/LocalRest.d
```diff
@@ -303,22 +303,53 @@ public void sleep (Duration timeout)
...
+    // Run a delegate after timeout, and until this.periodic is false
+    private void run ()
     {
         do
         {
             sleep(timeout);
             dg();
-        } while (periodic);
-    });
+        } while (this.periodic);
```
AndrejMitrovic: This should be:

```
do
{
    sleep(timeout);
    if (this.stopped)
        return;
    dg();
} while (this.periodic);
```

And the `stop` routine should do `this.stopped = false;`, so you will need to introduce a new `bool` variable.

This is because we need to have the ability to **cancel** a timer while it is sleeping.

So this PR needs to introduce two features:
- Stop a reoccurring timer (this.periodic is initially true)
- Cancel a one-shot timer before it runs (this.periodic is initially false)

AndrejMitrovic: I think this is why @Geod24 suggested that this could be handled by the Scheduler. Because the scheduler itself could know not to call `run()` if the timer has been cancelled.

That being said, we can improve upon the implementation later. I think the API will remain the same.

TrustHenry: I think so too.

TrustHenry:
Added periodic = false test in unittest as codecov error.

source/geod24/LocalRest.d
```diff
@@ -2015,16 +2058,23 @@ unittest
             return this.counter;
         }
```
AndrejMitrovic, line 2060: There is still no test for stopping a timer **before** it is called, not after it is called. I suggest doing this:

```diff
diff --git a/source/geod24/LocalRest.d b/source/geod24/LocalRest.d
index b2fa160..5c5efda 100644
--- a/source/geod24/LocalRest.d
+++ b/source/geod24/LocalRest.d
@@ -2031,6 +2031,7 @@ unittest
     static interface API
     {
         public void startTimer (bool periodic);
+        public void stopTimer ();
         public ulong getCounter ();
         public void resetCounter ();
     }
@@ -2045,6 +2046,11 @@ unittest
             this.timer = setTimer(100.msecs, &callback, periodic);
         }
 
+        public override void stopTimer ()
+        {
+            this.timer.stop();
+        }
+
         public void callback ()
         {
             this.counter++;
@@ -2073,8 +2079,9 @@ unittest
     assert(node.getCounter() == 3);
     node.resetCounter();
     node.startTimer(false);
+    node.stopTimer();
     core.thread.Thread.sleep(500.msecs);
-    assert(node.getCounter() == 1);
+    assert(node.getCounter() == 0);
     node.ctrl.shutdown();
 }
 ```

**Comment by AndrejMitrovic** [comment]
Fixed diff:

```diff
diff --git a/source/geod24/LocalRest.d b/source/geod24/LocalRest.d
index b2fa160..a62abbe 100644
--- a/source/geod24/LocalRest.d
+++ b/source/geod24/LocalRest.d
@@ -2031,6 +2031,7 @@ unittest
     static interface API
     {
         public void startTimer (bool periodic);
+        public void stopTimer ();
         public ulong getCounter ();
         public void resetCounter ();
     }
@@ -2045,6 +2046,11 @@ unittest
             this.timer = setTimer(100.msecs, &callback, periodic);
         }
 
+        public override void stopTimer ()
+        {
+            this.timer.stop();
+        }
+
         public void callback ()
         {
             this.counter++;
@@ -2072,9 +2078,10 @@ unittest
     // Check means the timer repeated and the timer stoped
     assert(node.getCounter() == 3);
     node.resetCounter();
+    node.stopTimer();
     node.startTimer(false);
     core.thread.Thread.sleep(500.msecs);
-    assert(node.getCounter() == 1);
+    assert(node.getCounter() == 0);
     node.ctrl.shutdown();
 }
 ```

File: source/geod24/LocalRest.d [hunk 11]
```diff
@@ -303,22 +303,59 @@ public void sleep (Duration timeout)
...
+
+/// Simple timer
+final class Timer
+{
+    private Duration timeout;
+    private void delegate () dg;
+    // Using a pointer so Timer may be copied and stop() will still work
+    private bool periodic;
+    // Whether to proceed with run delegate
+    private bool stopped;
```
```suggestion
    private Duration timeout;
    private void delegate () dg;
    // whether this timer is repeating
    private bool periodic;
    // whether this timer was stopped
    private bool stopped;
```

The `pointer` comment was outdated.

source/geod24/LocalRest.d
```diff
@@ -303,22 +303,59 @@ public void sleep (Duration timeout)
...
-    scheduler.schedule(
-    ()
+
+    Timer timer = new Timer(timeout, dg, periodic);
+    scheduler.schedule(() { timer.run(); });
+    return timer;
+}
+
+/// Simple timer
+final class Timer
```
```suggestion
public final class Timer
```

AndrejMitrovic (approved):
Just small ddoc issues.

TrustHenry:
Thank you.
All of @AndrejMitrovic suggestions have been applied.

source/geod24/LocalRest.d
```diff
@@ -303,22 +303,59 @@ public void sleep (Duration timeout)
...
+public Timer setTimer (Duration timeout, void delegate() dg,
     bool periodic = false)
 {
     assert(scheduler !is null, "Cannot call this delegate from the main thread");
     assert(dg !is null, "Cannot call this delegate if null");
-    scheduler.schedule(
-    ()
+
+    Timer timer = new Timer(timeout, dg, periodic);
+    scheduler.schedule(() { timer.run(); });
```
```suggestion
    scheduler.schedule(&timer.run);
```

This prevents an unneeded memory allocation.

Geod24 (commented):
I'd like the scheduler to be aware of the timer, but for the time being this'll do.

source/geod24/LocalRest.d
TrustHenry: I applied suggestion.

Issue #79, opened 2020-05-29
Return an error code when called with an non-existing method


Geod24:
Currently, if there is a mismatch between the node and interface, the node just abort:
https://github.com/Geod24/localrest/blob/bb15261bedbdd709dfe56c0288a6677b92290b9b/source/geod24/LocalRest.d#L440-L441

An example where this could happen:
```D
interface BaseAPI
{
    int getAnswer ();
}

interface AdvancedAPI : BaseAPI
{
    string getName ();
}

class SimpleNode : Base { override int getAnswer() { return 42; } }

// ...
unittest
{
    // `getNodeAddress` actually gets the address of a `SimpleNode`
    scope api = new RemoteAPI!Advanced(getNodeAddress());
    // This will abort the node and wait in main thread
    // It should throw an `Exception`, a la 404.
    assert(api.getName() == null);
}
```

Note that this is a synthetic test. In our use case, we have nodes communicating with each others, and they can be of two types, and switch *at runtime* between either types, so we need to keep a reference to them as a `AdvancedAPI` even if under the hood they might not be.

Geod24:
Fixed by #83

PR #83, opened 2020-06-26
Do not crash the node on unimplemented methods

Geod24:
There are legitimate use case for nodes that get called with the wrong method.
In the real world, this results in a 404 error being returned.
In Localrest it used to crash the remote node because of the assert, but no more.

### DIFF
--- a/source/geod24/LocalRest.d
+++ b/source/geod24/LocalRest.d
@@ -492,7 +492,9 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
                 }.format(member, ovrld.mangleof));
             }
         default:
-            assert(0, "Unmatched method name: " ~ cmd.method);
+            C.trySend(cmd.sender,
+                      Response(Status.ClientFailed, cmd.id,
+                               SerializedData("Method not found")));
         }
     }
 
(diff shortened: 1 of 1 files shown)


Geod24:
Fixes #79

source/geod24/LocalRest.d
```diff
@@ -475,7 +475,9 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
                 }.format(member, ovrld.mangleof));
             }
         default:
-            assert(0, "Unmatched method name: " ~ cmd.method);
+            C.trySend(cmd.sender,
+                      Response(Status.Failed, cmd.id,
+                               SerializedData("Method not implemented (404)")));
```
AndrejMitrovic: Would it be of any use to emit the method name in the message?

AndrejMitrovic: Hmm actually it would make even more sense to throw a custom 404 exception. If vibe.d does something similar we could handle both in our agora codebase.

Geod24:
Altered the code to throw a "client error" when there is one.

linked0 (approved):
LGTM.
I checked the modified code and approved it. I need to skim over the whole code though.

PR #84, opened 2020-07-09
Some minor refactoring and tweaks

Geod24:
I was working on fixing the issue we're seeing with `restart`, and those few things were in the way.


AndrejMitrovic (approved):
Looks good.

AndrejMitrovic:
There's some nothrow errors in there.

Geod24:
I'll let the attributes be inferred then, it's always done for nested functions anyway

AndrejMitrovic:
> it's always done for nested functions anyway

Oh I didn't know that. But actually that makes perfect sense.

PR #85, opened 2020-07-09
Add basic benchmarks

Geod24:
```
At the moment of writing, the mean is 16 micro seconds,
with spikes in the 40-60 range.
```

This has been on my laptop for a while.


AndrejMitrovic (approved):
LGTM.

PR #87, opened 2020-07-21
Add tests with DMD / LDC master

Geod24:


Geod24:
Or perhaps I should just switch to Github Actions to have cross-platform testing...

Geod24:
Well LDC still has the regression we caught: https://github.com/ldc-developers/ldc/issues/3501#issuecomment-661551187

Geod24:
Updated with a workaround for the Windows ICE we found (yay): https://github.com/ldc-developers/ldc/pull/3517#issuecomment-662654478

AndrejMitrovic:
> Ideally, these descriptors would be truly unique, but WinEH depends on string comparisons to check for matching catch clauses

what the..

source/geod24/LocalRest.d
```diff
@@ -103,6 +103,7 @@ module geod24.LocalRest;
 
 static import C = geod24.concurrency;
```
AndrejMitrovic, line 104: our of spawn => out of spawn

Geod24:
https://github.com/dlang/druntime/pull/2874

.github/workflows/main.yaml
```diff
@@ -0,0 +1,40 @@
...
+
+jobs:
+  main:
+    name: Run
+    # Matrix configuration
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ ubuntu-18.04, macOS-10.15, windows-2019 ]
+        dc: [ ldc-master, dmd-master, ldc-latest, dmd-latest, ldc-1.17.0, dmd-2.088.1 ]
```
AndrejMitrovic: How about gdc? We don't want to support it yet?

AndrejMitrovic: s/we/you

source/geod24/LocalRest.d
Geod24: What ? Where is this written ?

AndrejMitrovic: In the commit message. Sorry github makes it not easy to comment on the actual commit message.

.github/workflows/main.yaml
Geod24: Riiiiiight, I'll add it back

AndrejMitrovic:
> dlang/druntime#2874

I remember typeinfo in DLLs on Windows being a problem.. 10 years ago. :D

AndrejMitrovic:
I cannot reproduce the bug. :/

I tried DMD `2.091.0`, `2.092.1`, and ldc `1.21.0` on my Windows machine (cmd.exe not WSL). It seems maybe it's sensitive to the environment ~- maybe I have too much RAM (16GB)? I'm not sure how much is available on the CI.~

AndrejMitrovic:
Correction I only have 8GB of RAM, so that can't be it.

Geod24:
Done in #135

PR #88, opened 2020-07-21
Make sleep(Duration) nothrow

Geod24:


source/geod24/LocalRest.d
```diff
@@ -292,7 +292,7 @@ public void runTask (void delegate() dg)
 }
 
 /// Ditto
-public void sleep (Duration timeout)
+public void sleep (Duration timeout) nothrow
```
linked0, line 295: The reason why you make this function `nothrow` is just because it won't throw an exception? It's just a question about the purpose.

Geod24: Yes, and it allows callers to be `nothrow`. This function is just a thin wrapper around another function which is `nothrow`.

linked0 (approved):

PR #89, opened 2020-07-22
Fix restarting behavior (and out of range error)

AndrejMitrovic:


AndrejMitrovic:
First commit fails the CI. I'm not sure how to write it in a way that it only signals that it fails (it's captured by LocalRest).

source/geod24/LocalRest.d
```diff
@@ -2157,6 +2157,55 @@ unittest
...
+
+        public override void call0 ()
+        {
+            scope node2 = new RemoteAPI!API(node2Addr);
+            node2.call1();
+        }
+
+        public override void call1 ()
+        {
+            Thread.sleep(1.seconds);
```
Geod24: Use the `sleep` call, not `Thread.sleep`.
Also set a `__gshared bool hasRun` at the end so you can while / cas on it in the main thread, like I did with the network test yesterday (or just use a mutex with timeout...)

AndrejMitrovic: Hmm, I think I will still need to add a little bit of thread.sleep here though. Because I don't know at which point the Response will be received by node 1. Consider for example:

```diff
diff --git a/source/geod24/LocalRest.d b/source/geod24/LocalRest.d
index a06bd63..8343c9e 100644
--- a/source/geod24/LocalRest.d
+++ b/source/geod24/LocalRest.d
@@ -2168,6 +2168,7 @@ unittest
     }

     __gshared C.Tid node2Addr;
+    __gshared bool hasWokenUp;

     static class Node : API
     {
@@ -2181,13 +2182,14 @@ unittest

         public override void call1 ()
         {
-            Thread.sleep(1.seconds);
+            hasWokenUp = true;
         }
     }

     auto node1 = RemoteAPI!API.spawn!Node(500.msecs);
     auto node2 = RemoteAPI!API.spawn!Node();
     node2Addr = node2.ctrl.tid();
+    node2.ctrl.sleep(2.seconds, false);

     try
     {
@@ -2198,8 +2200,7 @@ unittest

     node1.ctrl.restart();

-    // this will trigger the range error
-    Thread.sleep(2.seconds);
+    while (!hasWokenUp) {}

     node1.ctrl.shutdown();
     node2.ctrl.shutdown();
```


After the `while` loop there, we still can't be sure the `Response` was sent back yet after `call1` returned.

Or did you have another idea?

AndrejMitrovic:
I figured out a solution via `schedule`.

source/geod24/LocalRest.d
```diff
@@ -2157,6 +2157,70 @@ unittest
...
+{
+    import core.atomic : atomicLoad, atomicStore;
+    static interface API
+    {
+        @safe:
+        public void call0 ();
+        public void call1 ();
+    }
+
+    __gshared C.Tid node1Addr;
```
AndrejMitrovic: Unused.

source/geod24/LocalRest.d
```diff
@@ -2157,6 +2157,68 @@ unittest
...
+    try
+    {
+        node1.call0();
+        assert(0, "This should have timed out");
+    }
+    catch (Exception e) {}
+
+    node1.ctrl.restart();
+
+    // after a while node 2 will receive a response to the timed-out request
```
AndrejMitrovic: I think this comment is wrong, it should say node 1.

source/geod24/LocalRest.d
```diff
@@ -110,16 +110,25 @@ import std.traits : fullyQualifiedName, Parameters, ReturnType;
...
 /// Data sent by the caller
 private struct Command
 {
     /// Tid of the sender thread (cannot be JSON serialized)
     C.Tid sender;
-    /// In order to support re-entrancy, every request contains an id
-    /// which should be copied in the `Response`
-    /// Initialized to `size_t.max` so not setting it crashes the program
-    size_t id = size_t.max;
+
```
AndrejMitrovic: Extra whitespace.

linked0 (approved):

source/geod24/LocalRest.d
```diff
@@ -2157,6 +2157,68 @@ unittest
...
+        public override void call0 ()
+        {
+            scope node2 = new RemoteAPI!API(node2Addr);
+            node2.call1();
+        }
+
+        public override void call1 ()
+        {
+            // when this event runs we know call1() has already returned
```
Geod24, line 2205: Smart

Geod24 (approved):
LGTM. Can you just invert the commit so we don't have a  commit failing tests ?

AndrejMitrovic:
Then it only makes sense to merge them I think?

AndrejMitrovic:
s/merge/squash

AndrejMitrovic:
Squashed.

source/geod24/LocalRest.d
Geod24: We can let it slide if you want 🙈

Geod24 (approved):
Merge at will

source/geod24/LocalRest.d
AndrejMitrovic: I fixed it though?

AndrejMitrovic: Oh no, it creeped back!!!

AndrejMitrovic:
I restored the *good* changes.

PR #90, opened 2020-07-30
Implement dependency injection mechanism

AndrejMitrovic:
@Geod24 this should simplify our test code in Agora.

### DIFF
--- a/source/geod24/LocalRest.d
+++ b/source/geod24/LocalRest.d
@@ -59,6 +59,12 @@
     might be processed / send by the node until the request is actually processed.
     There is also a `restart` method which accepts the same callback argument.
 
+    Inject:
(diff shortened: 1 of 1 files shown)


Geod24:
What's the benefit of doing this in LocalRest as opposed to what we're currently doing in Agora ?

AndrejMitrovic:
I started explaining it in a reply, but then I realized code speaks volumes. I'll write up a PR in Agora to show the difference.

AndrejMitrovic:
So this was useful but only in one single use-case, where we wanted to call a method which is not virtual and not exposed in the API. However it's not useful for getting values back since the call still happens in the node thread.

I tried to remove the need for `TestAPI` in our code but wasn't really successful.

I will close this for now.

PR #93, opened 2020-09-03
Change the InfoFiber stack size

TrustHenry:
This resolves the memory allocation fail issue in the GitHub Actions Windows-2019 OS Environment Test.
Stack size must be downsized on Windows due to overcommit memory.

### DIFF
--- a/source/geod24/concurrency.d
+++ b/source/geod24/concurrency.d
@@ -926,7 +926,7 @@ protected:
     {
         ThreadInfo info;
 
-        this(void delegate() op, size_t sz = 16 * 1024 * 1024) nothrow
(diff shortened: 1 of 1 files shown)


Geod24:
@AndrejMitrovic : Do you think that'll cut it ?

AndrejMitrovic:
Well, we previously increased the fiber size https://github.com/Geod24/localrest/pull/21/files because we ran out of stack space.

It's possible we're a little smarter about how we schedule requests now so stack overflows are less common. If this passes the Agora CI then I guess it's ok.

AndrejMitrovic:
It seems to work on my Windows machine. Let's go ahead with this then.

PR #94, opened 2020-09-25
Eliminate unnecessary scheduling of blocked Fibers

omerfirmak:
These changes eliminate unnecessary scheduling of blocked Fibers.

### DIFF
--- a/source/geod24/concurrency.d
+++ b/source/geod24/concurrency.d
@@ -925,6 +925,7 @@ protected:
     static class InfoFiber : Fiber
     {
         ThreadInfo info;
+        FiberCondition cond;
 
         this(void delegate() op, size_t sz = 512 * 1024) nothrow
         {
@@ -934,26 +935,30 @@ protected:
 
     protected class FiberCondition
     {
+        MonoTime limit = MonoTime.init;
+
+        private void registerToInfoFiber(InfoFiber info_fiber = cast(InfoFiber) Fiber.getThis()) nothrow
+        {
+            assert(info_fiber !is null, "This Fiber does not belong to FiberScheduler");
+            assert(info_fiber.sem is null, "This Fiber already has a registered FiberBinarySemaphore");
+            info_fiber.sem = this;
+        }
+
         void wait() nothrow
         {
+            this.registerToInfoFiber();
+            FiberScheduler.yield();
             scope (exit) notified = false;
-
-            while (!notified)
-                FiberScheduler.yield();
         }
 
         bool wait(Duration period) nothrow
         {
-            import core.time : MonoTime;
+            this.registerToInfoFiber();
+            limit = MonoTime.currTime + period;
+            FiberScheduler.yield();
+            limit = MonoTime.init;
 
             scope (exit) notified = false;
-
-            for (auto limit = MonoTime.currTime + period;
-                 !notified && !period.isNegative;
-                 period = limit - MonoTime.currTime)
-            {
-                FiberScheduler.yield();
-            }
             return notified;
         }
 
@@ -970,9 +975,28 @@ private:
     void dispatch()
     {
         import std.algorithm.mutation : remove;
-
         while (m_fibers.length > 0)
         {
+            // Is Fiber waiting on a FiberCondition?
+            if (m_fibers[m_pos].cond !is null)
+            {
+                // Is condition met?
+                // TRUE: Clear the condition and schedule the fiber
+                // FALSE: Skip it
+                if (m_fibers[m_pos].cond.notified ||
+                    (m_fibers[m_pos].cond.limit != MonoTime.init
+                        && MonoTime.currTime >= m_fibers[m_pos].cond.limit) )
+                {
+                    m_fibers[m_pos].cond = null;
+                }
+                else
+                {
+                    if (m_pos++ >= m_fibers.length - 1)
+                        m_pos = 0;
+                    continue;
+                }
+            }
+
             auto t = m_fibers[m_pos].call(Fiber.Rethrow.no);
             if (t !is null)
             {
@@ -991,7 +1015,7 @@ private:
     }
 
 private:
-    Fiber[] m_fibers;
+    InfoFiber[] m_fibers;
     size_t m_pos;
 }
 
--- a/source/geod24/concurrency.d
+++ b/source/geod24/concurrency.d
@@ -925,7 +925,9 @@ protected:
     static class InfoFiber : Fiber
     {
         ThreadInfo info;
-        FiberCondition cond;
+
+        /// Semaphore reference that this Fiber is blocked on
+        FiberBinarySemaphore sem;
 
         this(void delegate() op, size_t sz = 512 * 1024) nothrow
         {
@@ -933,42 +935,76 @@ protected:
         }
     }
 
-    protected class FiberCondition
+    protected class FiberBinarySemaphore
     {
-        MonoTime limit = MonoTime.init;
+
+        /***********************************************************************
+
+            Associate `FiberBinarySemaphore` with the running `Fiber`
+
+            `FiberScheduler` will check to see if the `Fiber` is blocking on a
+            `FiberBinarySemaphore` to avoid rescheduling it unnecessarily
+
+        ***********************************************************************/
 
         private void registerToInfoFiber(InfoFiber info_fiber = cast(InfoFiber) Fiber.getThis()) nothrow
         {
             assert(info_fiber !is null, "This Fiber does not belong to FiberScheduler");
             assert(info_fiber.sem is null, "This Fiber already has a registered FiberBinarySemaphore");
             info_fiber.sem = this;
+
         }
 
         void wait() nothrow
         {
             this.registerToInfoFiber();
             FiberScheduler.yield();
-            scope (exit) notified = false;
+            scope (exit) this.notified = false;
         }
 
         bool wait(Duration period) nothrow
         {
             this.registerToInfoFiber();
-            limit = MonoTime.currTime + period;
+            this.limit = MonoTime.currTime + period;
             FiberScheduler.yield();
-            limit = MonoTime.init;
 
-            scope (exit) notified = false;
-            return notified;
+            scope (exit)
+            {
+                this.limit = MonoTime.init;
+                this.notified = false;
+            }
+            return this.notified;
         }
 
         void notify() nothrow
         {
-            notified = true;
+            this.notified = true;
             FiberScheduler.yield();
         }
 
-        private bool notified;
+        /***********************************************************************
+
+            Query if `FiberBinarySemaphore` should still block
+
+            FiberBinarySemaphore will block the Fiber until it is notified or
+            the specified timeout is reached.
+
+        ***********************************************************************/
+
+        bool shouldBlock() nothrow
+        {
+            bool timed_out = (limit != MonoTime.init
+                                && MonoTime.currTime >= limit);
+
+            return !timed_out && !notified;
+        }
+
+    private:
+        /// Time limit that will eventually unblock the caller if a timeout is specified
+        MonoTime limit = MonoTime.init;
+
+        /// State of the semaphore
+        bool notified;
     }
 
 private:
@@ -977,23 +1013,21 @@ private:
         import std.algorithm.mutation : remove;
         while (m_fibers.length > 0)
         {
-            // Is Fiber waiting on a FiberCondition?
-            if (m_fibers[m_pos].cond !is null)
+            // Is Fiber waiting on a FiberBinarySemaphore?
+            if (auto sem = m_fibers[m_pos].sem)
             {
                 // Is condition met?
-                // TRUE: Clear the condition and schedule the fiber
+                // TRUE: Clear the sem and schedule the fiber
                 // FALSE: Skip it
-                if (m_fibers[m_pos].cond.notified ||
-                    (m_fibers[m_pos].cond.limit != MonoTime.init
-                        && MonoTime.currTime >= m_fibers[m_pos].cond.limit) )
-                {
-                    m_fibers[m_pos].cond = null;
-                }
-                else
+                if (sem.shouldBlock())
                 {
                     if (m_pos++ >= m_fibers.length - 1)
                         m_pos = 0;
                     continue;
+                }
+                else
+                {
+                    m_fibers[m_pos].sem = null;
                 }
             }
 
--- a/source/geod24/concurrency.d
+++ b/source/geod24/concurrency.d
@@ -831,8 +831,7 @@ class FiberScheduler
     {
         create(op);
         // Make sure the just-created fiber is run first
-        this.m_pos = this.m_fibers.length - 1;
-        dispatch();
+        dispatch(this.m_fibers.length - 1);
     }
 
     /**
@@ -1008,49 +1007,61 @@ protected:
     }
 
 private:
-    void dispatch()
+
+    /***********************************************************************
+
+        Start the scheduling loop
+
+        Param:
+            pos = m_fibers index of the Fiber to execute first
+
+    ***********************************************************************/
+
+    void dispatch(size_t pos = 0)
     {
         import std.algorithm.mutation : remove;
+
+        assert(pos < m_fibers.length);
         while (m_fibers.length > 0)
         {
             // Is Fiber waiting on a FiberBinarySemaphore?
-            if (auto sem = m_fibers[m_pos].sem)
+            if (auto sem = m_fibers[pos].sem)
             {
                 // Is condition met?
                 // TRUE: Clear the sem and schedule the fiber
                 // FALSE: Skip it
                 if (sem.shouldBlock())
                 {
-                    if (m_pos++ >= m_fibers.length - 1)
-                        m_pos = 0;
+                    if (pos++ >= m_fibers.length - 1)
+                        pos = 0;
                     continue;
                 }
                 else
                 {
-                    m_fibers[m_pos].sem = null;
+                    m_fibers[pos].sem = null;
                 }
             }
 
-            auto t = m_fibers[m_pos].call(Fiber.Rethrow.no);
+            auto t = m_fibers[pos].call(Fiber.Rethrow.no);
             if (t !is null)
             {
                 throw t;
             }
-            if (m_fibers[m_pos].state == Fiber.State.TERM)
+            if (m_fibers[pos].state == Fiber.State.TERM)
             {
-                if (m_pos >= (m_fibers = remove(m_fibers, m_pos)).length)
-                    m_pos = 0;
+                if (pos >= (m_fibers = remove(m_fibers, pos)).length)
+                    pos = 0;
             }
-            else if (m_pos++ >= m_fibers.length - 1)
+            else if (pos++ >= m_fibers.length - 1)
             {
-                m_pos = 0;
+                pos = 0;
             }
         }
     }
 
 private:
+    /// List of `InfoFiber`s currently in the system
     InfoFiber[] m_fibers;
-    size_t m_pos;
 }
 
 /// Ensure argument to `start` is run first
--- a/source/geod24/LocalRest.d
+++ b/source/geod24/LocalRest.d
@@ -218,8 +218,8 @@ private final class LocalScheduler : C.FiberScheduler
     import core.sync.condition;
     import core.sync.mutex;
 
-    /// Just a FiberCondition with a state
-    private struct Waiting { FiberCondition c; bool busy; }
+    /// Just a FiberBinarySemaphore with a state
+    private struct Waiting { FiberBinarySemaphore s; bool busy; }
 
     /// The 'Response' we are currently processing, if any
     private Response pending;
@@ -247,7 +247,7 @@ private final class LocalScheduler : C.FiberScheduler
     public Response waitResponse (ID id, Duration duration) nothrow
     {
         if (id !in this.waiting)
-            this.waiting[id] = Waiting(new FiberCondition, false);
+            this.waiting[id] = Waiting(new FiberBinarySemaphore, false);
 
         Waiting* ptr = &this.waiting[id];
         if (ptr.busy)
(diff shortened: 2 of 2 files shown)


source/geod24/concurrency.d
```diff
@@ -925,35 +925,45 @@ protected:
     static class InfoFiber : Fiber
     {
         ThreadInfo info;
+        FiberCondition cond;
 
         this(void delegate() op, size_t sz = 512 * 1024) nothrow
         {
+            cond = null;
```
```suggestion
```

The default is already `null`.

source/geod24/concurrency.d
```diff
@@ -925,35 +925,45 @@ protected:
...
             super(op, sz);
         }
     }
 
     protected class FiberCondition
     {
+        MonoTime limit = MonoTime.init;
```
AndrejMitrovic: Could use a ddoc-style comment with `///`.

source/geod24/concurrency.d
```diff
@@ -925,35 +925,45 @@ protected:
...
+
+        private void registerToInfoFiber() nothrow
+        {
+            /*
+            *  Register the FiberCondition to InfoFiber so that FiberScheduler can check if condition is met
+            */
```
```suggestion
            // Register the FiberCondition to InfoFiber so that FiberScheduler can check if condition is met
```

source/geod24/concurrency.d
AndrejMitrovic: Side-note: A lot of the existing fields are undocumented, they were taken verbatim from the Phobos implementation. But our style is to document all fields. It would be good to at least document the newly added ones and document the other ones later.

source/geod24/concurrency.d
```diff
@@ -925,35 +925,45 @@ protected:
...
+            auto infoF = cast(InfoFiber) Fiber.getThis();
+            assert(infoF !is null, "This Fiber does not belong to FiberScheduler");
+            assert(infoF.cond is null, "This Fiber already has a registered FiberCondition");
+            infoF.cond = this;
+        }
+
         void wait() nothrow
         {
+            registerToInfoFiber();
```
```suggestion
            this.registerToInfoFiber();
```

source/geod24/concurrency.d
```diff
@@ -925,35 +925,45 @@ protected:
...
             scope (exit) notified = false;
-
-            while (!notified)
-                FiberScheduler.yield();
         }
 
         bool wait(Duration period) nothrow
         {
-            import core.time : MonoTime;
+            registerToInfoFiber();
```
```suggestion
            this.registerToInfoFiber();
```

source/geod24/concurrency.d
```diff
@@ -925,35 +925,45 @@ protected:
...
+            limit = MonoTime.currTime + period;
```
```suggestion
            this.limit = MonoTime.currTime + period;
```

source/geod24/concurrency.d
```diff
@@ -925,35 +925,45 @@ protected:
...
+            FiberScheduler.yield();
+            limit = MonoTime.init;
```
AndrejMitrovic: Why is resetting the limit necessary? The `FiberCondition` is never re-used as far as I know. And it might hinder debugging if the fields are reset after use.

source/geod24/concurrency.d
```diff
@@ -925,35 +925,45 @@ protected:
...
+            cond = null;
             super(op, sz);
         }
     }
 
     protected class FiberCondition
     {
+        MonoTime limit = MonoTime.init;
+
+        private void registerToInfoFiber() nothrow
```
AndrejMitrovic: I wonder if you can move the registering to the `FiberCondition` constructor? That way we don't have to remember to call `registerToInfoFiber()` in all the `wait` routines. So basically we could turn `registerToInfoFiber` into a constructor and remove the calls to `registerToInfoFiber`.

source/geod24/concurrency.d
```diff
@@ -970,9 +980,30 @@ private:
...
         {
+            // Is Fiber waiting on a FiberCondition?
+            if (m_fibers[m_pos].cond !is null)
+            {
+                // Is condition met?
+                // TRUE: Clear the condition and schedule the fiber 
+                // FALSE: Skip it 
+                if (m_fibers[m_pos].cond.notified ||
+                    (m_fibers[m_pos].cond.limit != MonoTime.init 
+                        && MonoTime.currTime >= m_fibers[m_pos].cond.limit) )
```
AndrejMitrovic: A little D trick:

```suggestion
            if (auto cond = m_fibers[m_pos].cond)
            {
                // Is condition met?
                // TRUE: Clear the condition and schedule the fiber 
                // FALSE: Skip it 
                if (cond.notified || 
                    (cond.limit != MonoTime.init && MonoTime.currTime >= cond.limit))
```

source/geod24/concurrency.d
```diff
@@ -970,9 +980,30 @@ private:
...
+                {
+                    m_fibers[m_pos].cond = null;
+                }
+                else
+                {
+                    if (m_pos++ >= m_fibers.length - 1)
+                    {
+                        m_pos = 0;
+                    }
```
```suggestion
                    if (m_pos++ >= m_fibers.length - 1)
                        m_pos = 0;
```

source/geod24/concurrency.d
```diff
@@ -973,7 +972,17 @@ protected:
             FiberScheduler.yield();
         }
 
-        private bool notified;
+        bool should_block() nothrow
```
AndrejMitrovic: We use `camelCase`, not `snake_case` for function names. Also please add a ddoc comment.

source/geod24/concurrency.d
```diff
@@ -973,7 +972,17 @@ protected:
...
+        {
+            bool timedout = (limit != MonoTime.init
+                                && MonoTime.currTime >= limit);
+
+            return !timedout && !notified;
```
```suggestion
            bool timed_out = (limit != MonoTime.init
                                && MonoTime.currTime >= this.limit);

            return !timed_out && !notified;
```

source/geod24/concurrency.d
```diff
@@ -982,26 +991,24 @@ private:
         import std.algorithm.mutation : remove;
         while (m_fibers.length > 0)
         {
-            // Is Fiber waiting on a FiberCondition?
-            if (m_fibers[m_pos].cond !is null)
+            // Is Fiber waiting on a FiberBinarySemaphore?
+            if (m_fibers[m_pos].sem !is null)
```
AndrejMitrovic: As suggested in the previous commit, you can use `if (auto sem = m_fibers[m_pos].sem)` and then use `sem` knowing that it's non-null.

source/geod24/concurrency.d
```diff
@@ -982,26 +991,24 @@ private:
...
-                if (m_fibers[m_pos].cond.notified ||
-                    (m_fibers[m_pos].cond.limit != MonoTime.init 
-                        && MonoTime.currTime >= m_fibers[m_pos].cond.limit) )
-                {
-                    m_fibers[m_pos].cond = null;
-                }
-                else
+                // TRUE: Clear the sem and schedule the fiber
+                // FALSE: Skip it
+                if (m_fibers[m_pos].sem.should_block())
```
AndrejMitrovic: That's a good change!

AndrejMitrovic (changes requested):
Please make sure the commit header line does not exceed 80 characters. The `...` ellipsis can tell you if you're over the limit.

Also I highly suggest that the last commit is moved to a separate PR, because the first two commits are non-controversial and could be merged right away - as soon as some of the styling issues are fixed.

Good job!

source/geod24/concurrency.d
```diff
@@ -985,6 +986,161 @@ protected:
...
+    {
+        this(ulong size = 0)
+        {
+            this.size = size;
+            m_lock = new Mutex;
+        }
+
+        bool write(T val)
+        {
+            bool notify_peer, fiber_was_blocked = false;
```
AndrejMitrovic: By the way, this only sets `fiber_was_blocked` to `false` explicitly. It leaves `notify_peer` to equal `bool.init`. But `bool.init` is already false. I suggest moving each variable declaration to its own line (it follows our style guide), and we can avoid initializers when `Type.init` is already what we want.

```suggestion
            bool notify_peer;
```

Also the `fiber_was_blocked` declaration should be moved to where it's first used. This is also our coding style: we do not declare variables at the start of the function, but instead to the closest where they are initialized.

source/geod24/concurrency.d
```diff
@@ -985,6 +986,161 @@ protected:
...
+            }
+        }
+
+    private:
+        T[] m_buffer;
+        bool m_closed = false;
+        Mutex m_lock;
+
+        ChannelQueueEntry[] m_rxq;
+        ChannelQueueEntry[] m_txq;
```
AndrejMitrovic: These will need better names, it's not obvious what the `t` in `txq` means. I can understand that `m_rxq` could mean request queue. It's fine (and preferred) to use slightly longer descriptive variable names.

AndrejMitrovic: All of these fields will need documentation.

source/geod24/concurrency.d
```diff
@@ -985,6 +986,161 @@ protected:
         bool notified;
     }
 
+    class Channel (T)
```
AndrejMitrovic: Is the intention of this `Channel` implementation be used with multiple threads? It seems that way from the usage of Mutex. However, there are no multithreaded tests which use the channel in this PR. Only multi-fiber tests which use the same thread. So the locks aren't really being tested properly here AFAICT.

Also, if a Channel will be shared across multiple threads then it should likely be marked as `shared`. I'm cc'ing @Geod24 for input here.

source/geod24/concurrency.d
```diff
@@ -985,6 +986,161 @@ protected:
...
+
+                    *(qEnt.pVal) = val;
+                }
+                else if (m_buffer.length < size)
+                {
+                    m_buffer ~= val;
+                }
+                else
+                {
+                    qEnt = new ChannelQueueEntry(&val);
```
AndrejMitrovic: I think this is wrong. You are passing a pointer to a temporary.

AndrejMitrovic: It might be best to just keep a copy of the value in `ChannelQueueEntry` rather than a pointer.

source/geod24/concurrency.d
```diff
@@ -985,6 +986,161 @@ protected:
...
+            {
+                qEnt.sem.notify();
+            }
+            else if (qEnt)
+            {
+                fiber_was_blocked = true;
+                qEnt.sem.wait();
+            }
+
+            return !(fiber_was_blocked) || qEnt.success;
```
```suggestion
                qEnt.sem.wait();
                return false;  // fiber was blocked
            }

            returnq Ent.success;
```

It removes the need to have an additional state variable.

source/geod24/concurrency.d
```diff
@@ -985,6 +986,161 @@ protected:
...
+                notify_peer = m_rxq.length > 0;
+
+                if (m_closed)
+                {
+                    return false;
+                }
+                else if (m_rxq.length > 0)
+                {
+                    qEnt = m_rxq[0];
+                    m_rxq = m_rxq.remove(0);
```
```suggestion
                    m_rxq = m_rxq.popFront();
```

Can use range functions like this from `std.range`.

source/geod24/concurrency.d
```diff
@@ -985,6 +986,161 @@ protected:
...
+    private:
+        T[] m_buffer;
+        bool m_closed = false;
+        Mutex m_lock;
+
+        ChannelQueueEntry[] m_rxq;
+        ChannelQueueEntry[] m_txq;
+
+    public:
+        immutable ulong size;
```
AndrejMitrovic: This should have a better name / be documented well. For example, is this an upper bound on the channel size, IOW is it a limit? Does the size mean the number of bytes, or the number of items? And what happens when the size limit is reached?

source/geod24/concurrency.d
```diff
@@ -985,6 +986,161 @@ protected:
         bool notified;
     }
 
+    class Channel (T)
+    {
+        this(ulong size = 0)
```
AndrejMitrovic: I'm guessing if size = 0 it changes the behavior of the channel. Buffered vs unbuffered. This needs to be documented.

And in fact I think having a default value is probably not a good idea.

source/geod24/concurrency.d
AndrejMitrovic: But actually using an array as a queue is inefficient. It's going to cause a lot of reallocations, because the array keeps growing but its pointer also keeps being shifted to the right. It's better to use something like a `DList` from `std.container` instead.

source/geod24/concurrency.d
```diff
@@ -985,6 +986,161 @@ protected:
...
+                    return false;
+                }
+                else if (m_rxq.length > 0)
+                {
+                    qEnt = m_rxq[0];
+                    m_rxq = m_rxq.remove(0);
+
+                    *(qEnt.pVal) = val;
+                }
+                else if (m_buffer.length < size)
```
AndrejMitrovic: This seems to have a nice property of also handling the case where `size` is zero, but it's much better to be explicit about the intention.

```suggestion
                else if (this.size > 0 && m_buffer.length < this.size)
```

AndrejMitrovic: And again I suggest renaming it. To something like `size_limit` or `max_size`, etc.

AndrejMitrovic: I also suggest to add small comments, something like:

```
if (...)  // buffered
{
}
else if (...)  // unbuffered
{
}
```

source/geod24/concurrency.d
```diff
@@ -985,6 +986,161 @@ protected:
...
+
+                    if (m_txq.length > 0)
+                    {
+                        qEnt = m_txq[0];
+                        m_txq = m_txq.remove(0);
+
+                        m_buffer ~= *(qEnt.pVal);
+                    }
+                }
+                else if (m_closed)
```
AndrejMitrovic: How come this isn't the first check that we do, but only the second one? The buffer must be flushed?

AndrejMitrovic (changes requested):
Completed the first review round of Channels. I'll have more comments after I get some feedback from you. Thanks.

source/geod24/concurrency.d
omerfirmak: Old habits :) Noted.

source/geod24/concurrency.d
omerfirmak: Yes, my intention was Channels to be thread safe but only to be used between Fibers. The tests are lacking, will add more.

source/geod24/concurrency.d
omerfirmak: qEnt can be a scoped variable but I think it needs to be a reference. 
At this point, we know that the current fiber will block. If a fiber closes this channel, it will modify the `success` field of the `ChannelQueueEntry` and that needs to be visible to the blocked fiber. Is there a better Dlang way of doing this?

source/geod24/concurrency.d
omerfirmak: Getting blocked doesn't always mean a failure. This doesn't cover the case where the caller gets blocked and eventually get woken up by a read/write. In that case we would need to return true.

source/geod24/concurrency.d
omerfirmak: Yes, I am aware the storage is not optimal. I will check `DList` out. Thanks!

source/geod24/concurrency.d
omerfirmak: Yes you are correct. I gave it a default value to be close to go channel syntax. Should we require caller to explicitly give 0 or have an overload for it?

source/geod24/concurrency.d
omerfirmak: Current implementation lets already buffered messages to be read after the channel is closed. Do we have any preference in that?

source/geod24/concurrency.d
omerfirmak: Oh, I misunderstood what you meant. The local `val` will not go out of scope as the Fiber will block until pointer to `val` inside `ChannelQueueEntry` is processed. Having a `T*` eliminates the need of extra copy to and from `qEnt` for both reference and value types

omerfirmak: In fact, to minimize the copying, I tried to make `val` a `auto ref T` but that didn't compile. Compiler strangely says that "auto can only be used as part of auto ref for template function parameters" which is exactly what I am trying to do.

AndrejMitrovic: Oh so it waits indefinitely until a `read` call processes this `ChannelQueueEntry`?

AndrejMitrovic: By the way, to make it possible to use `auto ref` you need to make the function a template. Something like: `bool write()(auto ref T val)`.

source/geod24/concurrency.d
AndrejMitrovic: It's ok to keep it as-is if you already plan to improve the storage later. I just wanted to make a suggestion for the future. :)

source/geod24/concurrency.d
AndrejMitrovic: > go channel syntax

Alright, I wasn't aware. A sensible default is also ok, as long as it's documented as such.

source/geod24/concurrency.d
AndrejMitrovic: Hmm, I think we don't have a preference. What does golang do?

source/geod24/concurrency.d
omerfirmak: `notified` gets reset after `wait` , I simply followed the same pattern. If we will rely on them not being reused we should have an assertion for it I guess.

source/geod24/concurrency.d
omerfirmak: That would be problematic, if the Fiber creates a `FiberBinarySemaphore` and yields before calling `wait` for whatever reason. `FiberScheduler` will still think that it is waiting on the semaphore.

source/geod24/concurrency.d
omerfirmak: Yes that is correct, they will be blocked until a corresponding `read` comes or channel is `close`'d.

source/geod24/concurrency.d
omerfirmak: Just tested it, Golang lets existing messages to be read after the channel gets closed.

source/geod24/concurrency.d
```diff
@@ -925,35 +925,45 @@ protected:
...
+            /*
+            *  Register the FiberCondition to InfoFiber so that FiberScheduler can check if condition is met
+            */
+            auto infoF = cast(InfoFiber) Fiber.getThis();
+            assert(infoF !is null, "This Fiber does not belong to FiberScheduler");
+            assert(infoF.cond is null, "This Fiber already has a registered FiberCondition");
+            infoF.cond = this;
+        }
+
         void wait() nothrow
```
Geod24, line 957: Could we pass the `InfoFiber` as an argument to avoid relying on global state in `registerToInfoFiber`.
Not to mentioned, we would be able to inline it inside `wait`.

source/geod24/concurrency.d
Geod24: I have a preference towards letting messages be read. So you can send an "end notification" and close immediately, instead of having to send an "end notification" and wait for an ACK before closing the channel (or have the other party close it).

source/geod24/concurrency.d
Geod24: Default is unbuffered and I think that's sane. It means you always context switch when writing, and context switch on reading if there's no message. If the channel is buffered, you might, or might not, context switch on write.

source/geod24/concurrency.d
omerfirmak: Who would be supplying the `InfoFiber` ? 

Relying on `Fiber.getThis()` helps with a certain scenario. `registerToInfoFiber` binds the current Fiber and itself at the time of `wait` call. This enables multiple Fibers to wait on same condition, if we ever need that. I made it into a function to reduce duplication since we have overloaded `wait`. We can also just drop `wait()` and have a single `wait (Duration period = 0)` to handle the infinite wait case as well.

source/geod24/concurrency.d
omerfirmak: @Geod24 , what is your take on this?

omerfirmak:
I removed the Channel commit. All comments are resolved for the first 2 commits apart from the one about resetting semaphore fields.

source/geod24/concurrency.d
Geod24: I think those variables should be reset when entering the function. In another review comment you mentioned that we could keep the possibility of multiple fibers waiting on the same condition, but that would not work if we reset `notified` at the end of the function here.

omerfirmak: Oh yes, multiple fibers waiting on the same condition would not work. 

There is another issue with resetting them at the start though. Imagine creating a condition and passing a reference of it to another thread. If condition gets signaled before you wait on it, you will miss that event and may even end up waiting indefinitely. I am not sure if this can be solved with additional locking, but I doubt it.

I think we should keep it as-is.

Geod24: Alright, let's keep it as is for now

Geod24:
Regarding the last commit: I think `m_pos` needs to be a member, because if we want direct hands off semantic, we need to update the scheduler position without jumping back inside the dispatch function, don't we ?

source/geod24/concurrency.d
Geod24: > Who would be supplying the `InfoFiber` ?

As a first step, you could just make it a default argument.

> This enables multiple Fibers to wait on same condition, if we ever need that.

So, here's my 2cts on it: Fibers are thread-local and *never* moved across threads. Each thread would have one or more scheduler (e.g. a task running in a scheduler might have its own scheduler as a task pool). A thread should *only* block if all its fibers are blocked. Otherwise it can schedule non-blocked fiber.
So there is a possibility that fibers block on the same condition, but that might be handled at a higher level, and fibers are likely to have their own condition.

> We can also just drop wait() and have a single `wait (Duration period = 0)` to handle the infinite wait case as well.

That'd be sweet. Maybe another PR though.

omerfirmak:
> Regarding the last commit: I think `m_pos` needs to be a member, because if we want direct hands off semantic, we need to update the scheduler position without jumping back inside the dispatch function, don't we ?

I don't think so. If we somehow directly `call()` the blocked Fiber1 from the running Fiber2, we should still end up back at the Fiber2 once Fiber1 yields. Only when Fiber2 yields, we would go back to the dispatch loop. There should be no reason to change m_pos if I am not mistaken.

source/geod24/concurrency.d
omerfirmak: > Each thread would have one or more scheduler (e.g. a task running in a scheduler might have its own scheduler as a task pool).

That should be possible with the current implementation.

> A thread should only block if all its fibers are blocked. Otherwise it can schedule non-blocked fiber.

That would be the ideal case, yes. But, if we want that level of control, I suppose we would need a fiber version of mutex, condvar etc, a more complex scheduler and run everything strictly in fibers.

PR #95, opened 2020-10-07
Implement Go-like concurrency

omerfirmak:
Solves #41
Solves #40 
Solves #49

### DIFF
--- a/source/geod24/LocalRest.d
+++ b/source/geod24/LocalRest.d
@@ -590,7 +590,7 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
                                 control.filter = filter_api;
                                 break;
                             default:
-                                assert(0);
+                                assert(0, "Got invalid Variant.Type: " ~ msg.tag);
                         }
                     }
 
--- a/source/geod24/LocalRest.d
+++ b/source/geod24/LocalRest.d
@@ -231,6 +231,11 @@ public void runTask (void delegate() dg) nothrow
 public void sleep (Duration timeout) nothrow
 {
     assert(!isMainThread(), "Cannot call this function from the main thread");
+
+    // Duration.init (0.seconds) is infinite timeout, ignore
+    if (timeout == Duration.init)
+        return;
+
     scope sem = scheduler.new FiberBinarySemaphore();
     sem.wait(timeout);
 }
--- a/source/geod24/LocalRest.d
+++ b/source/geod24/LocalRest.d
@@ -179,9 +179,6 @@ private struct ArgWrapper (T...)
 /// We need a scheduler to simulate an event loop and to be re-entrant
 private C.FiberScheduler scheduler;
 
-/// Whether this is the main thread
-private bool is_main_thread;
-
 /***********************************************************************
 
     Check if the current context is running inside the main thread and
@@ -196,11 +193,8 @@ private bool isMainThread() nothrow
 {
     // we are in the main thread
     if (scheduler is null)
-    {
         scheduler = new C.FiberScheduler;
-        is_main_thread = true;
-    }
-    return is_main_thread;
+    return C.thisScheduler() is null;
 }
 
 /*******************************************************************************
--- a/source/geod24/LocalRest.d
+++ b/source/geod24/LocalRest.d
@@ -579,6 +579,8 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
                                 if (e.callback !is null)
                                     e.callback(node);
                                 exc.restart = e.restart;
+                                if (!e.restart)
+                                    chn.close();
                                 throw exc;
                             case Variant.Type.timeCommand:
                                 TimeCommand s = msg.time;
@@ -621,7 +623,6 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
                         break;
                 }
             }
-            chn.close();
         }
         catch (Throwable t)
         {
--- a/source/geod24/LocalRest.d
+++ b/source/geod24/LocalRest.d
@@ -236,8 +236,8 @@ public void sleep (Duration timeout) nothrow
     if (timeout == Duration.init)
         return;
 
-    scope sem = scheduler.new FiberBinarySemaphore();
-    sem.wait(timeout);
+    scope blocker = scheduler.new FiberBlocker();
+    blocker.wait(timeout);
 }
 
 /*******************************************************************************
--- a/source/geod24/LocalRest.d
+++ b/source/geod24/LocalRest.d
@@ -698,6 +698,23 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
 
         /***********************************************************************
 
+            Send `Variant` to peer in a Fiber
+
+            Param:
+                var = `Variant` to send
+
+        ***********************************************************************/
+
+        private void sendToPeer (Variant var)
+        {
+            if (isMainThread())
+                scheduler.start({ this.listen_chn.write(var); });
+            else
+                this.listen_chn.write(var);
+        }
+
+        /***********************************************************************
+
             Send an async message to the thread to immediately shut down.
 
             Params:
@@ -710,10 +727,7 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
         public void shutdown (void function (API) callback = null)
             @trusted
         {
-            if (isMainThread())
-                scheduler.start({ this.listen_chn.write(Variant(ShutdownCommand!API(callback, false))); });
-            else
-                this.listen_chn.write(Variant(ShutdownCommand!API(callback, false)));
+            sendToPeer(Variant(ShutdownCommand!API(callback, false)));
         }
 
         /***********************************************************************
@@ -733,10 +747,7 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
         public void restart (void function (API) callback = null)
             @trusted
         {
-            if (isMainThread())
-                scheduler.start({ this.listen_chn.write(Variant(ShutdownCommand!API(callback, true))); });
-            else
-                this.listen_chn.write(Variant(ShutdownCommand!API(callback, true)));
+            sendToPeer(Variant(ShutdownCommand!API(callback, true)));
         }
 
         /***********************************************************************
@@ -757,10 +768,7 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
 
         public void sleep (Duration d, bool dropMessages = false) @trusted
         {
-            if (isMainThread())
-                scheduler.start({ this.listen_chn.write(Variant(TimeCommand(d, dropMessages))); });
-            else
-                this.listen_chn.write(Variant(TimeCommand(d, dropMessages)));
+            sendToPeer(Variant(TimeCommand(d, dropMessages)));
         }
 
         /***********************************************************************
@@ -839,10 +847,7 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
                 enum mangled = getBestMatch!Overloads;
             }
 
-            if (isMainThread())
-                scheduler.start({ this.listen_chn.write(Variant(FilterAPI(mangled, pretty))); });
-            else
-                this.listen_chn.write(Variant(FilterAPI(mangled, pretty)));
+            sendToPeer(Variant(FilterAPI(mangled, pretty)));
         }
 
 
@@ -854,10 +859,7 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
 
         public void clearFilter () @trusted
         {
-            if (isMainThread())
-                scheduler.start({ this.listen_chn.write(Variant(FilterAPI(""))); });
-            else
-                this.listen_chn.write(Variant(FilterAPI("")));
+            sendToPeer(Variant(FilterAPI("")));
         }
 
         /***********************************************************************
--- a/source/geod24/LocalRest.d
+++ b/source/geod24/LocalRest.d
@@ -576,11 +576,11 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
                                 break;
                             case Variant.Type.shutdownCommand:
                                 ShutdownCommand!API e = msg.shutdown;
+                                if (!e.restart)
+                                    chn.close();
                                 if (e.callback !is null)
                                     e.callback(node);
                                 exc.restart = e.restart;
-                                if (!e.restart)
-                                    chn.close();
                                 throw exc;
                             case Variant.Type.timeCommand:
                                 TimeCommand s = msg.time;
--- a/source/geod24/LocalRest.d
+++ b/source/geod24/LocalRest.d
@@ -110,6 +110,17 @@ import std.traits : fullyQualifiedName, Parameters, ReturnType;
 import core.thread;
 import core.time;
 
+/// Data sent by the caller
+private struct Command
+{
+    /// Sequence id of Command
+    size_t id;
+    /// Method to call
+    string method;
+    /// Serialized arguments to the method
+    SerializedData args;
+}
+
 /// Ask the node to exhibit a certain behavior for a given time
 private struct TimeCommand
 {
@@ -117,6 +128,16 @@ private struct TimeCommand
     Duration dur;
     /// Whether or not affected messages should be dropped
     bool drop = false;
+}
+
+/// Ask the node to shut down
+private struct ShutdownCommand
+{
+    /// Any callback to call before the Node's destructor is called
+    void function (Object) callback;
+
+    /// Whether we're restarting or really shutting down
+    bool restart;
 }
 
 /// Filter out requests before they reach a node
@@ -150,6 +171,8 @@ private struct Response
 {
     /// Final status of a request (failed, timeout, success, etc)
     Status status;
+    /// Response id
+    size_t id;
     /// If `status == Status.Success`, the serialized return value.
     /// Otherwise, it contains `Exception.toString()`.
     SerializedData data;
@@ -176,6 +199,196 @@ private struct ArgWrapper (T...)
     T args;
 }
 
+// very simple & limited variant, to keep it performant.
+// should be replaced by a real Variant later
+private struct Variant
+{
+    this (Command msg) { this.cmd = msg; this.tag = Variant.Type.command; }
+    this (Response msg) { this.res = msg; this.tag = Variant.Type.response; }
+    this (FilterAPI msg) { this.filter = msg; this.tag = Variant.Type.filter; }
+    this (TimeCommand msg) { this.time = msg; this.tag = Variant.Type.timeCommand; }
+    this (ShutdownCommand msg) { this.shutdown = msg; this.tag = Variant.Type.shutdownCommand; }
+
+    union
+    {
+        Command cmd;
+        Response res;
+        FilterAPI filter;
+        TimeCommand time;
+        ShutdownCommand shutdown;
+    }
+
+    Type tag;
+
+    /// Type of a request
+    enum Type : ubyte
+    {
+        command,
+        response,
+        filter,
+        timeCommand,
+        shutdownCommand,
+    }
+}
+
+private alias CommandChn = C.Channel!Variant;
+private alias RespChn = C.Channel!Response;
+
+/// Represents a connection between a server and a client
+private class Connection
+{
+    ///
+    this () @trusted nothrow
+    {
+        this.command_chn = new CommandChn();
+        this.resp_chn = new RespChn();
+    }
+
+    /*******************************************************************************
+
+        Send a message over the Connection
+
+        Params:
+            T = Type of the message, should be support by the `Variant` type
+            msg = Message to be sent
+
+        Returns:
+            Success/failure
+
+    *******************************************************************************/
+
+    bool sendCommand (T) (T msg) @trusted
+    {
+        bool ret;
+
+        if (isMainThread())
+            scheduler.start({ ret = this.command_chn.write(Variant(msg)); });
+        else
+            ret = this.command_chn.write(Variant(msg));
+
+        return ret;
+    }
+
+    /*******************************************************************************
+
+        Get a unique id for a `Command` to be sent from this `Connection`
+
+        Returns:
+            Unique Command id
+
+    *******************************************************************************/
+
+    size_t getNextCommandId () @safe
+    {
+        return this.next_cmd_id++;
+    }
+
+    /*******************************************************************************
+
+        Wait for a `Response` with specific id
+
+        Params:
+            resp_id = Response id to wait for
+            timeout = optional timeout duration for the operation
+
+        Returns:
+            Response
+
+    *******************************************************************************/
+
+    Response waitResponse (size_t resp_id, Duration timeout = Duration.init) @trusted
+    {
+        if (isMainThread())
+        {
+            Response res;
+            scheduler.start({
+                // Loop until we get the Response we are looking for
+                while (this.resp_chn.read(res, timeout))
+                    if (res.id == resp_id)
+                        return;
+                res = Response(Status.Timeout, resp_id, SerializedData("Request timed-out"));
+            });
+            return res;
+        }
+        else
+        {
+            // Response may already be ready
+            if (auto existing_res = (resp_id in this.waiting_list))
+                return existing_res.res;
+
+            // Block while waiting the Response
+            auto blocker = C.thisScheduler().new FiberBlocker();
+            this.waiting_list[resp_id] = Waiting(blocker);
+            if (blocker.wait(timeout))
+                return this.waiting_list[resp_id].res;
+            else
+                return Response(Status.Timeout, resp_id, SerializedData("Request timed-out"));
+        }
+    }
+
+    /*******************************************************************************
+
+        Notify the task waiting for a Response
+
+        Params:
+            res = Newly arrived Response
+
+    *******************************************************************************/
+
+    void notifyWaiter (Response res)
+    {
+        // Is the waiting Fiber already blocked?
+        if (auto waiter = (res.id in this.waiting_list))
+        {
+            waiter.res = res;
+            waiter.blocker.notify();
+        }
+        else // Fiber is not yet blocked, create an entry for the related Fiber to use
+            this.waiting_list[res.id] = Waiting(null, res);
+    }
+
+    /*******************************************************************************
+
+        Close the `Channel`s associated with this `Connection`. Blocked `waitResponse`
+        calls will timeout and blocked `sendCommand` calls will fail
+
+    *******************************************************************************/
+
+    void close ()
+    {
+        this.command_chn.close();
+        this.resp_chn.close();
+    }
+
+    ///
+    private struct Waiting
+    {
+        C.FiberScheduler.FiberBlocker blocker;
+        Response res;
+    }
+
+    /// List of Fibers waiting for a Response from this Connection
+    private Waiting[size_t] waiting_list;
+
+    /// Next Command ID
+    private size_t next_cmd_id;
+
+    /// Channel to send `Command`s to
+    private CommandChn command_chn;
+
+    /// Channel to read `Response`s from
+    private RespChn resp_chn;
+}
+
+/// `Channel` type that the nodes will listen for new `Connection`s
+public alias BindChn = C.Channel!Connection;
+
+/// Thread local outgoing `Connection` list
+private Connection[RespChn] outgoing_conns;
+
+/// Thread local incoming `Connection` list
+private Connection[CommandChn] incoming_conns;
+
 /// We need a scheduler to simulate an event loop and to be re-entrant
 private C.FiberScheduler scheduler;
 
@@ -189,7 +402,7 @@ private C.FiberScheduler scheduler;
 
 ***********************************************************************/
 
-private bool isMainThread() nothrow
+private bool isMainThread () @trusted nothrow
 {
     // we are in the main thread
     if (scheduler is null)
@@ -330,61 +543,6 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
 {
     static assert (!serializerInvalidReason!(S).length, serializerInvalidReason!S);
 
-    alias RestChannel = C.Channel!Variant;
-
-    /// Ask the node to shut down
-    private struct ShutdownCommand (API)
-    {
-        /// Any callback to call before the Node's destructor is called
-        void function (API) callback;
-
-        /// Whether we're restarting or really shutting down
-        bool restart;
-    }
-
-    /// Data sent by the caller
-    private struct Command
-    {
-        /// Channel to be used when responding to this Command
-        RestChannel resp_chn;
-        /// Method to call
-        string method;
-        /// Serialized arguments to the method
-        SerializedData args;
-    }
-
-    // very simple & limited variant, to keep it performant.
-    // should be replaced by a real Variant later
-    static struct Variant
-    {
-        this (Command msg) { this.cmd = msg; this.tag = Variant.Type.command; }
-        this (Response msg) { this.res = msg; this.tag = Variant.Type.response; }
-        this (FilterAPI msg) { this.filter = msg; this.tag = Variant.Type.filter; }
-        this (TimeCommand msg) { this.time = msg; this.tag = Variant.Type.timeCommand; }
-        this (ShutdownCommand!API msg) { this.shutdown = msg; this.tag = Variant.Type.shutdownCommand; }
-
-        union
-        {
-            Command cmd;
-            Response res;
-            FilterAPI filter;
-            TimeCommand time;
-            ShutdownCommand!API shutdown;
-        }
-
-        Type tag;
-
-        /// Type of a request
-        enum Type : ubyte
-        {
-            command,
-            response,
-            filter,
-            timeCommand,
-            shutdownCommand,
-        }
-    }
-
     /***************************************************************************
 
         Instantiate a node and start it
@@ -415,7 +573,7 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
         CtorParams!Impl args, Duration timeout = Duration.init,
         string file = __FILE__, int line = __LINE__)
     {
-        auto chn = new RestChannel(16);
+        auto chn = new BindChn();
         new Thread(
         {
             spawned!(Impl)(chn, file, line, args);
@@ -443,10 +601,11 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
             cmd    = the command to run (contains the method name and the arguments)
             node   = the node to invoke the method on
             filter = used for filtering API calls (returns default response)
+            resp_chn = `Channel` to send the `Response` to
 
     ***************************************************************************/
 
-    private static void handleCommand (Command cmd, API node, FilterAPI filter)
+    private static void handleCommand (Command cmd, API node, FilterAPI filter, RespChn resp_chn)
     {
         switch (cmd.method)
         {
@@ -456,7 +615,7 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
                 mixin(
                 q{
                     case `%2$s`:
-                    Response res = Response(Status.Failed);
+                    Response res = Response(Status.Failed, cmd.id);
 
                     // Provide informative message in case of filtered method
                     if (cmd.method == filter.func_mangleof)
@@ -481,14 +640,12 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
                         }
                     }
 
-                    cmd.resp_chn.write(Variant(res));
-                    cmd.resp_chn.close();
+                    resp_chn.write(res);
                     return;
                 }.format(member, ovrld.mangleof));
             }
         default:
-            cmd.resp_chn.write(Variant(Response(Status.ClientFailed, SerializedData("Method not found"))));
-            cmd.resp_chn.close();
+            resp_chn.write(Response(Status.ClientFailed, cmd.id, SerializedData("Method not found")));
             break;
         }
     }
@@ -503,7 +660,7 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
 
        Params:
            Implementation = Type of the implementation to instantiate
-           chn = The channel on which to "listen" to receive new "connections"
+           bind_chn = The channel on which to "listen" to receive new "connections"
            file = Path to the file that spawned this node
            line = Line number in the `file` that spawned this node
            cargs = Arguments to `Implementation`'s constructor
@@ -511,7 +668,7 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
     ***************************************************************************/
 
     private static void spawned (Implementation) (
-        RestChannel chn, string file, int line, CtorParams!Implementation cargs)
+        BindChn bind_chn, string file, int line, CtorParams!Implementation cargs)
         nothrow
     {
         import std.datetime.systime : Clock, SysTime;
@@ -553,31 +710,83 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
             // Control the node behavior
             Control control;
 
+            struct AwaitingMessage
+            {
+                /// Message
+                Variant var;
+                /// Originating `Connection`
+                Connection conn;
+            }
             // we need to keep track of messages which were ignored when
             // node.sleep() was used, and then handle each message in sequence.
-            Variant[] await_msgs;
+            AwaitingMessage[] await_msgs;
 
             scheduler.start(() {
-                while (1)
+                C.SelectEntry[] read_list;
+                C.SelectEntry[] write_list;
+
+                while (true)
                 {
+                    Connection new_conn;
+                    Response res;
                     Variant msg;
-                    if (chn.read(msg, 10.msecs))
+
+                    read_list.length = 0;
+                    assumeSafeAppend(read_list);
+                    write_list.length = 0;
+                    assumeSafeAppend(write_list);
+
+                    foreach (ref conn; incoming_conns)
+                        if (!conn.command_chn.isClosed())
+                            read_list ~= C.SelectEntry(conn.command_chn, &msg);
+                    foreach (ref conn; outgoing_conns)
+                        if (!conn.resp_chn.isClosed())
+                            read_list ~= C.SelectEntry(conn.resp_chn, &res);
+                    read_list ~= C.SelectEntry(bind_chn, &new_conn);
+
+                    auto sel_ret = C.select(read_list, write_list, 10.msecs);
+
+                    if (!control.isSleeping())
                     {
+                        foreach (ref await_msg; await_msgs)
+                            handleCommand(await_msg.var.cmd, node, control.filter, await_msg.conn.resp_chn);
+                        await_msgs.length = 0;
+                        assumeSafeAppend(await_msgs);
+                    }
+
+                    if (!sel_ret.success)
+                        continue;
+
+                    // Bind chn
+                    if (cast(BindChn) read_list[sel_ret.id].selectable)
+                    {
+                        incoming_conns[new_conn.command_chn] = new_conn;
+                    }
+                    // Command
+                    else if (auto comm_chn = cast(CommandChn) read_list[sel_ret.id].selectable)
+                    {
+                        auto curr_conn = incoming_conns[comm_chn];
                         switch (msg.tag)
                         {
                             case Variant.Type.command:
                                 Command cmd = msg.cmd;
                                 if (!control.isSleeping())
                                     scheduler.spawn({
-                                        handleCommand(cmd, node, control.filter);
+                                        handleCommand(cmd, node, control.filter, curr_conn.resp_chn);
                                     });
                                 else if (!control.drop)
-                                    await_msgs ~= msg;
+                                    await_msgs ~= AwaitingMessage(msg, curr_conn);
                                 break;
                             case Variant.Type.shutdownCommand:
-                                ShutdownCommand!API e = msg.shutdown;
+                                ShutdownCommand e = msg.shutdown;
                                 if (!e.restart)
-                                    chn.close();
+                                {
+                                    bind_chn.close();
+                                    incoming_conns.each!((conn) => conn.close());
+                                }
+                                outgoing_conns.each!((conn) => conn.close());
+                                outgoing_conns.clear();
+
                                 if (e.callback !is null)
                                     e.callback(node);
                                 exc.restart = e.restart;
@@ -595,17 +804,10 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
                                 assert(0, "Got invalid Variant.Type: " ~ msg.tag);
                         }
                     }
-
-                    // now handle any leftover messages after any sleep() call
-                    if (!control.isSleeping())
+                    else if (auto resp_chn = cast(RespChn) read_list[sel_ret.id].selectable)
                     {
-                        foreach (ref await_msg; await_msgs)
-                        {
-                            if (await_msg.tag == Variant.Type.command)
-                                handleCommand(await_msg.cmd, node, control.filter);
-                        }
-                        await_msgs.length = 0;
-                        assumeSafeAppend(await_msgs);
+                        // Response
+                        outgoing_conns[resp_chn].notifyWaiter(res);
                     }
                 }
             });
@@ -642,7 +844,10 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
     private const Duration timeout;
 
     /// Main Channel that this Node will listen for incoming messages
-    private RestChannel listen_chn;
+    private BindChn bind_chn;
+
+    /// Connection between this instance and the node main thread
+    private Connection conn;
 
     /***************************************************************************
 
@@ -652,18 +857,28 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
         In order to instantiate a node, see the static `spawn` function.
 
         Params:
-          listen_chn = `RestChannel` of the node.
+          bind_chn = `BindChn` of the node.
                 This can usually be obtained by `geod24.concurrency.locate`.
           timeout = any timeout to use
 
     ***************************************************************************/
 
-    public this (Object listen_chn, Duration timeout = Duration.init)
-        @safe @nogc pure nothrow
+    public this (BindChn bind_chn, Duration timeout = Duration.init) @trusted nothrow
     {
-        this.listen_chn = cast(RestChannel) listen_chn;
+        import std.exception : assumeWontThrow;
+
+        this.bind_chn = bind_chn;
         this.timeout = timeout;
-        assert(this.listen_chn);
+        assert(bind_chn);
+
+        // Create a new Connection, register it and send it to the peer
+        // TODO: What to do when bind_chn is closed?
+        this.conn = new Connection();
+        outgoing_conns[this.conn.resp_chn] = this.conn;
+        if (isMainThread())
+            assumeWontThrow(scheduler.start({ bind_chn.write(this.conn); }));
+        else
+            bind_chn.write(this.conn);
     }
 
     /***************************************************************************
@@ -684,34 +899,16 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
     {
         /***********************************************************************
 
-            Returns the `Tid` this `RemoteAPI` wraps
+            Returns the `BindChn` main thread of this `RemoteAPI` listens to
+            for new `Connections`
 
             This can be useful for calling `geod24.concurrency.register` or similar.
-            Note that the `Tid` should not be used directly, as our event loop,
-            would error out on an unknown message.
 
         ***********************************************************************/
 
-        public RestChannel chn () @nogc pure nothrow
+        public BindChn chn () @nogc pure nothrow
         {
-            return this.listen_chn;
-        }
-
-        /***********************************************************************
-
-            Send `Variant` to peer in a Fiber
-
-            Param:
-                var = `Variant` to send
-
-        ***********************************************************************/
-
-        private void sendToPeer (Variant var)
-        {
-            if (isMainThread())
-                scheduler.start({ this.listen_chn.write(var); });
-            else
-                this.listen_chn.write(var);
+            return this.bind_chn;
         }
 
         /***********************************************************************
@@ -725,10 +922,10 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
 
         ***********************************************************************/
 
-        public void shutdown (void function (API) callback = null)
+        public void shutdown (void function (Object) callback = null)
             @trusted
         {
-            sendToPeer(Variant(ShutdownCommand!API(callback, false)));
+            this.conn.sendCommand(ShutdownCommand(callback, false));
         }
 
         /***********************************************************************
@@ -745,19 +942,15 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
 
         ***********************************************************************/
 
-        public void restart (void function (API) callback = null)
+        public void restart (void function (Object) callback = null)
             @trusted
         {
-            sendToPeer(Variant(ShutdownCommand!API(callback, true)));
+            this.conn.sendCommand(ShutdownCommand(callback, true));
         }
 
         /***********************************************************************
 
             Make the remote node sleep for `Duration`
-
-            The remote node will call `Thread.sleep`, becoming completely
-            unresponsive, potentially having multiple tasks hanging.
-            This is useful to simulate a delay or a network outage.
 
             Params:
               delay = Duration the node will sleep for
@@ -769,7 +962,7 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
 
         public void sleep (Duration d, bool dropMessages = false) @trusted
         {
-            sendToPeer(Variant(TimeCommand(d, dropMessages)));
+            this.conn.sendCommand(TimeCommand(d, dropMessages));
         }
 
         /***********************************************************************
@@ -848,7 +1041,7 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
                 enum mangled = getBestMatch!Overloads;
             }
 
-            sendToPeer(Variant(FilterAPI(mangled, pretty)));
+            this.conn.sendCommand(FilterAPI(mangled, pretty));
         }
 
 
@@ -860,7 +1053,7 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
 
         public void clearFilter () @trusted
         {
-            sendToPeer(Variant(FilterAPI("")));
+            this.conn.sendCommand(FilterAPI(""));
         }
 
         /***********************************************************************
@@ -883,7 +1076,7 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
 
         public void withTimeout (Dg) (Duration timeout, scope Dg dg)
         {
-            scope api = new RemoteAPI(this.listen_chn, timeout);
+            scope api = new RemoteAPI(this.bind_chn, timeout);
             static assert(is(typeof({ dg(api); })),
                           "Provided argument of type `" ~ Dg.stringof ~
                           "` is not callable with argument type `scope " ~
@@ -912,28 +1105,11 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
                     // this overload needs to be
                     auto res = () @trusted {
                         auto serialized = S.serialize(ArgWrapper!(Parameters!ovrld)(params));
-                        auto resp_chn = new RestChannel(4);
-                        auto command = Command(resp_chn, ovrld.mangleof, SerializedData(serialized));
+                        auto command = Command(this.conn.getNextCommandId(), ovrld.mangleof, SerializedData(serialized));
 
-                        Response sendCommandAndWaitResponse ()
-                        {
-                            Variant var;
-                            if(!this.listen_chn.write(Variant(command)))
-                                throw new Exception("Connection with peer closed");
-                            if (resp_chn.read(var, this.timeout))
-                                return var.res;
-                            else
-                                return Response(Status.Timeout, SerializedData("Request timed-out"));
-                        }
-
-                        Response res;
-                        // for the main thread, we run the "event loop" until
-                        // the request we're interested in receives a response.
-                        if (isMainThread())
-                            scheduler.start({ res = sendCommandAndWaitResponse(); });
-                        else
-                            res = sendCommandAndWaitResponse();
-                        return res;
+                        if(!this.conn.sendCommand(command))
+                            throw new Exception("Connection with peer closed");
+                        return this.conn.waitResponse(command.id, this.timeout);
                     }();
 
                     if (res.status == Status.Failed)
@@ -1007,7 +1183,7 @@ unittest
         ~this () { dtor_called = true; }
     }
 
-    static void onDestroy (API node)
+    static void onDestroy (Object node)
     {
         assert(!dtor_called);
         auto mock = cast(MockAPI)node;
@@ -1147,7 +1323,7 @@ unittest
     static class SlaveNode : API
     {
         @safe:
-        this(Object masterChn)
+        this(BindChn masterChn)
         {
             this.master = new RemoteAPI!API(masterChn);
         }
@@ -1199,7 +1375,7 @@ unittest
 {
     static import geod24.concurrency;
 
-    __gshared Object[string] tbn;
+    __gshared BindChn[string] tbn;
 
     static interface API
     {
@@ -1329,7 +1505,7 @@ unittest
 // Simulate temporary outage
 unittest
 {
-    __gshared Object n1chn;
+    __gshared BindChn n1chn;
 
     static interface API
     {
@@ -1400,8 +1576,7 @@ unittest
 // Filter commands
 unittest
 {
-    __gshared Object node_chn;
-
+    __gshared BindChn node_chn;
     static interface API
     {
         size_t fooCount();
@@ -1422,11 +1597,6 @@ unittest
         size_t bar_count;
         RemoteAPI!API remote;
 
-        public this()
-        {
-            this.remote = new RemoteAPI!API(node_chn);
-        }
-
         override size_t fooCount() { return this.foo_count; }
         override size_t fooIntCount() { return this.foo_int_count; }
         override size_t barCount() { return this.bar_count; }
@@ -1439,6 +1609,9 @@ unittest
 
         override void callFoo()
         {
+            if (!this.remote)
+                this.remote = new RemoteAPI!API(node_chn);
+
             try
             {
                 this.remote.foo();
@@ -1451,6 +1624,9 @@ unittest
 
         override void callFoo(int arg)
         {
+            if (!this.remote)
+                this.remote = new RemoteAPI!API(node_chn);
+
             try
             {
                 this.remote.foo(arg);
@@ -1463,6 +1639,9 @@ unittest
 
         override void callBar(int arg)
         {
+            if (!this.remote)
+                this.remote = new RemoteAPI!API(node_chn);
+
             try
             {
                 this.remote.bar(arg);
@@ -1587,7 +1766,7 @@ unittest
     assert(called);
 
     // Test that attributes are inferred based on the delegate
-    void doTest () @safe pure nothrow @nogc
+    void doTest () @safe nothrow
     {
         called = false;
         node.ctrl.withTimeout(Duration.zero,
@@ -1663,7 +1842,7 @@ unittest
     static import geod24.concurrency;
     import std.exception;
 
-    __gshared Object node_chn;
+    __gshared BindChn node_chn;
 
     static interface API
     {
@@ -1704,7 +1883,7 @@ unittest
     static import geod24.concurrency;
     import std.exception;
 
-    __gshared Object node_chn;
+    __gshared BindChn node_chn;
 
     static interface API
     {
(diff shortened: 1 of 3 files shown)


AndrejMitrovic:
Do you think it would make sense to split up the buffered and non-buffered implementation into separate classes with a common interface? Granted this would introduce virtual calls, and you couldn't use `() (auto ref T)` then (because templates aren't virtual).

It's just that the code is a bit complex to follow when the class has multiple modes of operation.

source/geod24/concurrency.d
```diff
@@ -1006,6 +1007,601 @@ protected:
...
+
+            Buffered mode:
+
+                If a reader is already blocked on the Channel, writer copies the
+                message to reader's buffer and wakes up the reader by yielding
+
+                If no reader is in the wait queue and the buffer is not full
+                writer puts the message in the Channel buffer and returns immediatly
+
+                If no reader is in the wait queue and the buffer is full
```
AndrejMitrovic: > and the buffer is full

I don't see a check for the buffer length inside of `write()`. Is this implicit?

source/geod24/concurrency.d
```diff
@@ -1006,6 +1007,601 @@ protected:
...
+            regardless of the operation mode
+
+            Param:
+                val = Message to write to channel
+            Returns:
+                Success/Failure
+
+        ***********************************************************************/
+
+        bool write () (auto ref T val)
```
AndrejMitrovic: What if an Exception is thrown? None of the methods are marked `nothrow`, so it's possible the call site would have to be wrapped in both a try/catch and a check for the return value.

In particular can we recover from an exception being thrown?

source/geod24/concurrency.d
```diff
@@ -1006,6 +1007,601 @@ protected:
...
+                message to reader's buffer and wakes up the reader by yielding
+
+                If no reader is in the wait queue and the buffer is not full
+                writer puts the message in the Channel buffer and returns immediatly
+
+                If no reader is in the wait queue and the buffer is full
+                writer appends itself to write wait queue and blocks
+
+            If channel is closed, it returns immediatly with a failure,
+            regardless of the operation mode
```
AndrejMitrovic: This needs to clarify if it returns false **only** if the channel is closed, or if there are other cases where write can return false.

source/geod24/concurrency.d
```diff
@@ -1006,6 +1007,601 @@ protected:
...
+
+            Param:
+                val = Message to write to channel
+            Returns:
+                Success/Failure
+
+        ***********************************************************************/
+
+        bool write () (auto ref T val)
+        {
```
AndrejMitrovic, line 1199: How come there is no lock in here? There doesn't need to be?

source/geod24/concurrency.d
```diff
@@ -1006,6 +1007,601 @@ protected:
...
+                sel_state.sem.notify();
+            }
+            else if (!sel_state.isConsumed())
+            {
+                // enqueueEntry will unlock the channel
+                enqueueEntry(this.writeq, val, sel_state, sel_id);
+            }
+            else // Non-blocking operation failed and sel_state is consumed
+            {
+                this.lock.unlock();
```
AndrejMitrovic: This is a bit strange to me because I don't see where the lock is initially acquired (it's outside this method maybe).

source/geod24/concurrency.d
```diff
@@ -1006,6 +1007,601 @@ protected:
...
+
+        SelectState
+
+        Hold the state of a group of selectRead/selectWrite calls. Synchronizes
+        peers that will consume those calls, so that only one
+        selectRead/selectWrite call is completed.
+
+    ***********************************************************************/
+
+    class SelectState
```
AndrejMitrovic: I recommend marking inner classes static in order to avoid the addition of unnecessary context pointers (it makes the class more expensive to allocate).

In addition if the class is not supposed to be derived from, mark it `final`. It should help with de-virtualizing methods (all public class methods in D are virtual).

```suggestion
    final static class SelectState
```

source/geod24/concurrency.d
```diff
@@ -1006,6 +1007,601 @@ protected:
...
+    /***********************************************************************
+
+        A golang style channel implementation with buffered and unbuffered
+        operation modes
+
+        Intended to be used between Fibers
+
+    ***********************************************************************/
+
+    class Channel (T)
```
```suggestion
    final static class Channel (T)
```

omerfirmak:
> Do you think it would make sense to split up the buffered and non-buffered implementation into separate classes with a common interface? Granted this would introduce virtual calls, and you couldn't use `() (auto ref T)` then (because templates aren't virtual).
> 
> It's just that the code is a bit complex to follow when the class has multiple modes of operation.

A standalone unbuffered channel would definitely be simpler, but a standalone buffered implementation might not be any simpler than the current hybrid implementation we have. We don't have anything extra in there to make it also work unbuffered. Unbuffered channels are just buffered channels with a full buffer, they come for free when you implement buffered ones :) 

Most of the complexity comes from the select operations. I originally wanted them to be on a completely separate layer, but baking them into the channel itself made more sense later.

source/geod24/concurrency.d
```diff
@@ -1006,6 +1007,601 @@ protected:
...
+            Param:
+                val = Message to write to channel
+                sel_state = SelectState instace of the select call being executed
+                sel_id = id of the select call being executed
+
+        ***********************************************************************/
+
+        void selectWrite() (T* val, SelectState sel_state, uint sel_id)
+        {
+            bool success = tryWrite(*val, sel_state, sel_id);
```
```suggestion
            assert(val !is null);
            assert(sel_state !is null);
            bool success = tryWrite(*val, sel_state, sel_id);
```

Alternatively:

```
void selectWrite() (T* val, SelectState sel_state, uint sel_id)
in
{
    assert(val !is null);
    assert(sel_state !is null);
}
body
{
}
```

I think since recently you can also write `do` instead of `body`.

source/geod24/concurrency.d
```diff
@@ -1006,6 +1007,601 @@ protected:
...
+            }
+            else if (!sel_state.isConsumed())
+            {
+                // enqueueEntry will unlock the channel
+                enqueueEntry(this.writeq, val, sel_state, sel_id);
+            }
+            else // Non-blocking operation failed and sel_state is consumed
+            {
+                this.lock.unlock();
+                if (sel_state.id == -1)
```
AndrejMitrovic: Please don't do this. You are comparing an unsigned type `uint` with `-1`.

AndrejMitrovic: I also don't see where it's ever initialized to `-1` / `uint.max`.

source/geod24/concurrency.d
```diff
@@ -1006,6 +1007,601 @@ protected:
...
+            if (this.buffer.length > 0)
+            {
+                // if dequeueEntry fails, we will try to consume caller_sel again.
+                if (!caller_sel || write_ent || caller_sel.tryConsume(caller_sel_id))
+                {
+                    output = this.buffer[0];
+                    this.buffer = this.buffer.remove(0);
+
+                    if (write_ent)
+                    {
```
```suggestion
                    {
                        assert(write_ent.pVal !is null);
```

source/geod24/concurrency.d
```diff
@@ -1006,6 +1007,601 @@ protected:
...
+                }
+                else
+                {
+                    return false;
+                }
+            }
+            // if dequeueEntry returns a valid entry, it always successfully consumes the related select states.
+            // the race between 2 select calls is resolved in dequeueEntry.
+            else if (write_ent)
+            {
```
```suggestion
            {
                assert(write_ent.pVal !is null);
```

source/geod24/concurrency.d
```diff
@@ -1006,6 +1007,601 @@ protected:
...
+            }
+            else if (!sel_state.isConsumed())
+            {
+                enqueueEntry(this.readq, val, sel_state, sel_id);
+            }
+            else
+            {
+                this.lock.unlock();
+
+                if (sel_state.id == -1)
```
AndrejMitrovic: Same comment about comparing unsigned type to `-1`.

source/geod24/concurrency.d
```diff
@@ -1006,6 +1007,601 @@ protected:
...
+                            newly created ChannelQueueEntry
+                sel_id = id of the select call creating the new ChannelQueueEntry
+            Return:
+                newly created ChannelQueueEntry
+
+        ***********************************************************************/
+
+        private ChannelQueueEntry enqueueEntry(ref ChannelQueueEntry[] entryq, T* pVal, SelectState sel_state = null, uint sel_id = 0)
+        {
+            ChannelQueueEntry q_ent = new ChannelQueueEntry(pVal, sel_state, sel_id);
```
AndrejMitrovic: Why does a `ChannelQueueEntry` need to be a class? It seems like a lot of overhead. It might be better to just use a struct and avoid unnecessary GC allocations.

AndrejMitrovic (commented):
First wave of reviews. I'll have more coming later.

source/geod24/concurrency.d
omerfirmak: Actually there can't be any readers in the queue with a non-empty buffer. That could have been worded better, I will update it.

source/geod24/concurrency.d
omerfirmak: There is no reason for them to throw Exceptions. I will mark them all `nothrow`.

source/geod24/concurrency.d
omerfirmak: `tryWrite` acquires the lock. It releases the lock if it successfully completes a non-blocking operation.
In case of a blocking operation; `tryWrite` returns with the mutex still locked and `enqueueEntry` unlocks when it is finished.

AndrejMitrovic: Ah yes. That makes sense, thanks!

source/geod24/concurrency.d
omerfirmak: `tryRead`/`tryWrite` are always the first ones to acquire the locks.

source/geod24/concurrency.d
omerfirmak: I wish there was a compiler warning for this, will make it an `int`. 
`id` can be set to `-1` in `dequeueEntry`

AndrejMitrovic: > I wish there was a compiler warning for this

We do, too. But Walter Bright is against it. :P

source/geod24/concurrency.d
omerfirmak: Maybe this should be done at the enqueue stage to better localize the source.

source/geod24/concurrency.d
omerfirmak: I needed it to be a heap-allocated-reference-type and went for a class.
Would this function do any allocations other than the one we do explicitly here? I assumed that only the reference of the `q_ent` gets passed around. Wouldn't a heap allocated struct still use GC?

AndrejMitrovic: Ah if you need reference semantics, it's ok to keep it as a class. I didn't realize that was the use-case.

AndrejMitrovic: > Wouldn't a heap allocated struct still use GC?

Yes. I wrongly assumed we can use value semantics here.

source/geod24/concurrency.d
omerfirmak: It is the only case, added an explanation for that.

source/geod24/concurrency.d
omerfirmak: I pushed a commit for this, could you check if that is idiomatic D?

omerfirmak:
> First wave of reviews. I'll have more coming later.

Thank you for the review! I pushed separate commits for each comment. We can squash them once we are happy with the changes.

source/geod24/concurrency.d
AndrejMitrovic: It looks good!

source/geod24/concurrency.d
```diff
@@ -897,6 +898,78 @@ class FiberScheduler
...
+            assert(info_fiber.sem is null, "This Fiber already has a registered FiberBinarySemaphore");
+            info_fiber.sem = this;
+
+        }
+
+        void wait() nothrow
+        {
+            this.registerToInfoFiber();
+            FiberScheduler.yield();
+            scope (exit) this.notified = false;
```
AndrejMitrovic: I think this is in the wrong position. It should be placed at the start of the wait function.

AndrejMitrovic: Hold on, actually the function is nothrow. So you can just remove `scope (exit)` entirely.

source/geod24/concurrency.d
```diff
@@ -934,76 +1007,616 @@ protected:
...
+                message to reader's buffer and wakes up the reader by yielding
+
+                If no reader is ready in the wait queue, writer appends itself
+                to write wait queue and blocks
+
+            Buffered mode:
+
+                If a reader is already blocked on the Channel, writer copies the
+                message to reader's buffer and wakes up the reader by yielding.
+                After execution comes back to the writer, it returns immediatly.
```
Geod24: Technically this differs from the Golang semantic. In Go, writing to an unbuffered channel *always* block the caller (either really blocking, or hand off to pending reader), but writing to a buffered channel never blocks for the writer if the channel isn't full.

Example code:
```Go
package main

import (
	"fmt"
)

func receiver(ch chan string) {
	foo := <- ch
	fmt.Println("Received: ", foo)
}

func main() {
         // Create an unbuffered channel
	ch := make(chan string)
	// Create a buffered channel
        ch := make(chan string, 2)
	go receiver(ch)
	ch <- "World"
	ch <- "World"	 // Remove me if using unbuffered channel
	fmt.Println("Hello, playground")
}
```

If you run it on the playground, [you can see](https://play.golang.org/p/ZqcJvc1pCsk) that it only prints "Hello playground", never prints the "Hello" sent.

Now, the title of the issue mentioned "Go style concurrency" not because we need to emulate Go in every aspect, but because we wanted to provide a directional, pipe based API. So while we don't have to emulate every aspects of it, I wonder if this  semantic is actually relied on somewhere, and how useful it actually is.

Thinking about it, it would make sense to have a buffered channel when your writer has high throughput, and unbuffered channel when the reader does (because of the hands off semantic).

source/geod24/concurrency.d
```diff
@@ -897,6 +898,78 @@ class FiberScheduler
...
+
+        private void registerToInfoFiber(InfoFiber info_fiber = cast(InfoFiber) Fiber.getThis()) nothrow
+        {
+            assert(info_fiber !is null, "This Fiber does not belong to FiberScheduler");
+            assert(info_fiber.sem is null, "This Fiber already has a registered FiberBinarySemaphore");
+            info_fiber.sem = this;
+
+        }
+
+        void wait() nothrow
```
Geod24: Style nitpick: We always put space between a parenthesis and a function declaration or definition.
So this should be:
```suggestion
        void wait () nothrow
```

On the other hand, if you call it, no space:
```D
foo.wait();
```

This makes grepping / Ctrl + F for the definition super simple.

source/geod24/concurrency.d
```diff
@@ -934,76 +1007,616 @@ protected:
...
+        private bool tryRead(ref T output, SelectState caller_sel = null, int caller_sel_id = 0) nothrow
+        {
+            this.lock.lock_nothrow();
+
+            ChannelQueueEntry write_ent = dequeueEntry(this.writeq, caller_sel, caller_sel_id);
+
+            if (this.buffer.length > 0)
+            {
+                // if dequeueEntry fails, we will try to consume caller_sel again.
+                if (!caller_sel || write_ent || caller_sel.tryConsume(caller_sel_id))
```
AndrejMitrovic: Won't this segfault if `caller_sel` is null and `write_ent` is true?

source/geod24/concurrency.d
omerfirmak: This is solely because the `FiberBinarySemaphore` yields the running Fiber on `notify()` calls. I tried to change that before, but the agora tests failed.

source/geod24/concurrency.d
```diff
@@ -934,76 +1007,616 @@ protected:
         }
     }
 
-    protected class FiberBinarySemaphore
+    /***********************************************************************
+
+        SelectState
```
Geod24: Did you read [the documentation specs](https://dlang.org/spec/ddoc.html) ?
If you want some example of how things would look like, I recommend you look at the agora docs.
In particular, [a simple module like ECC](https://bpfkorea.github.io/agora/agora/common/crypto/ECC.html) would produce [this documentation](https://bpfkorea.github.io/agora/agora/common/crypto/ECC.html).

You can see what looks readable and what doesn't, e.g. if you go to the `Scalar` struct and check the functions, you'll see which ones are well documented, which one is not, and which one should have its description split in two ;)

source/geod24/concurrency.d
omerfirmak: If `caller_sel` is null, the latter 2 expressions won't get evaluated.

AndrejMitrovic: Oh right, I misread it! Personally I would make this a bit clearer:

```suggestion
                if (write_ent || (caller_sel && caller_sel.tryConsume(caller_sel_id))
```

I think that would have the same semantics?

Geod24:
I will give this a test, but I don't see any handling of `shared` vs non-`shared` data, which might wreak havoc in the design if not considered from early on.

source/geod24/concurrency.d
omerfirmak: I didn't :) Will take a look now.

source/geod24/concurrency.d
omerfirmak: Noted.

source/geod24/concurrency.d
omerfirmak: We still want to execute the body of `if` when `caller_sel` is null. I think this would fail in that case, when `write_ent` is null.

AndrejMitrovic: Ah yep. Carry on. :)

source/geod24/concurrency.d
omerfirmak: Yes, I will remove it. I know the order of scope statements matter when there is multiple. They get executed in reverse. But does it matter when there is only one?

omerfirmak:
> I will give this a test, but I don't see any handling of `shared` vs non-`shared` data, which might wreak havoc in the design if not considered from early on.

Please be gentle :)

source/geod24/concurrency.d
omerfirmak: Applied this style to functions I recently touched or wrote myself.

omerfirmak:
@AndrejMitrovic @Geod24  What do you think about the `select()` implementation I just pushed? Especially about the `void*` and pointer casts. Should we have a protection against misuse?

btw, I didn't bother with doc and styling just yet.

AndrejMitrovic:
> What do you think about the select() implementation I just pushed?

I'm not sure. What is the benefit?

omerfirmak:
> > What do you think about the select() implementation I just pushed?
> 
> I'm not sure. What is the benefit?

Mainly, it lets user to programmatically select on multiple channels without hard coding `selectRead`/`selectWrite` calls. It also lets us manage the select flow and hide some implementation details. For example, `selectRead`/`selectWrite` can fail (because of a low probability race condition) but we can make sure `select()` does not by looping until we succeed. A `select()` with a timeout will need some special handling, which we can do correctly in `select()` and not rely on user to get it right every time.

source/geod24/concurrency.d
```diff
@@ -990,150 +966,889 @@ protected:
...
+            this.lock.unlock_nothrow();
+        }
+
+        /***********************************************************************
+
+            Return the length of the internal buffer
+
+        ***********************************************************************/
+
+        ulong length () nothrow
```
Geod24:
```suggestion
        size_t length () nothrow
```

Geod24:
Wow, great work! It's fairly hard to assess the impact at the moment, but this is definitely going in the right direction. Kudos for the care you put into documenting it, highly appreciated.

source/geod24/concurrency.d
omerfirmak: This is closer to Golang behavior with the new FiberScheduler in this [branch](https://github.com/omerfirmak/localrest/tree/go-style-sched).

omerfirmak:
I have broken the code into 5 commits. We can go commit by commit and merge them all at once or have them each on a seperate PR. But it makes more sense to me to merge them all at once. @Geod24 @AndrejMitrovic

source/geod24/array.d
```diff
@@ -0,0 +1,315 @@
+/**
```
AndrejMitrovic: I think this module should be part of its own commit.

source/geod24/concurrency.d
```diff
@@ -823,6 +861,7 @@ public class InfoThread : Thread
  */
 class FiberScheduler
 {
+
```
AndrejMitrovic, line 888: boop

source/geod24/concurrency.d
```diff
@@ -815,6 +820,39 @@ public class InfoThread : Thread
...
+/***************************************************************************
+
+    Set current running FiberScheduler
+
+    Params:
+        value = Reference to the current FiberScheduler
+
+***************************************************************************/
+
+public void thisScheduler (FiberScheduler value) nothrow
```
AndrejMitrovic: I guess you could mark this as a `@property` and then use it like `thisScheduler = null` for example. Because `thisScheduler(arg)` looks really strange.

source/geod24/concurrency.d
```diff
@@ -946,39 +985,52 @@ protected:
...
+        private void registerToInfoFiber (InfoFiber info_fiber = cast(InfoFiber) Fiber.getThis()) nothrow
         {
             assert(info_fiber !is null, "This Fiber does not belong to FiberScheduler");
             assert(info_fiber.sem is null, "This Fiber already has a registered FiberBinarySemaphore");
             info_fiber.sem = this;
 
         }
 
-        void wait() nothrow
+        void wait () nothrow
```
AndrejMitrovic: In the future please make these styling changes in their own commit.

source/geod24/concurrency.d
```diff
@@ -1057,6 +1111,8 @@ private:
                 pos = 0;
             }
         }
+
+        thisScheduler(null);
```
AndrejMitrovic: Not in a `scope (exit)` at the beginning? I guess it doesn't matter after an exception is thrown..

source/geod24/concurrency.d
```diff
@@ -1446,3 +1502,665 @@ private:
...
+    A golang style channel implementation with buffered and unbuffered
+    operation modes
+
+    Intended to be used between Fibers
+
+***********************************************************************/
+
+final public class Channel (T)
+{
+
```
AndrejMitrovic: boop

source/geod24/concurrency.d
```diff
@@ -1446,3 +1502,665 @@ private:
...
+/***********************************************************************
+
+    A golang style channel implementation with buffered and unbuffered
+    operation modes
+
+    Intended to be used between Fibers
+
+***********************************************************************/
+
+final public class Channel (T)
```
AndrejMitrovic: `T` needs `Parameter` documentation. Are there any constraints here? T can be anything?

source/geod24/concurrency.d
```diff
@@ -1446,3 +1502,665 @@ private:
...
+{
+
+    /***********************************************************************
+
+        Constructs a Channel
+
+        Param:
+            max_size = Maximum amount of T a Channel can buffer
+                            (0 -> Unbuffered operation,
```
AndrejMitrovic: Unindent please so the text aligns.

source/geod24/concurrency.d
```diff
@@ -1446,3 +1502,665 @@ private:
...
+        tryRead reads a message if it is possible to do so without blocking.
+
+        If the tryRead is being executed in a select call, it tries to consume the
+        `caller_sel` with the given `caller_sel_id`. It only proceeds if it can
+        successfully consume the `caller_sel`
+
+        Param:
+            output = Field to write the message to
+            caller_sel = SelectState instace of the select call being executed
+            caller_sel_id = id of the select call being executed
```
AndrejMitrovic: This parameter documentation is wrong.

source/geod24/concurrency.d
```diff
@@ -1446,3 +1502,665 @@ private:
...
+        ChannelQueueEntry write_ent = dequeueEntry(this.writeq);
+
+        if (this.buffer.length > 0)
+        {
+            output = this.buffer.front();
+            this.buffer.popFront();
+
+            if (write_ent)
+            {
+                this.buffer.putBack(*(write_ent.pVal));
```
AndrejMitrovic: `this.buffer.putBack(*write_ent.pVal)` works fine too.

source/geod24/concurrency.d
```diff
@@ -1446,3 +1502,665 @@ private:
...
+            }
+        }
+        // if dequeueEntry returns a valid entry, it always successfully consumes the related select states.
+        // the race between 2 select calls is resolved in dequeueEntry.
+        else if (write_ent)
+        {
+            output = *(write_ent.pVal);
```
AndrejMitrovic: output = *write_ent.pVal;

source/geod24/concurrency.d
```diff
@@ -1446,3 +1502,665 @@ private:
...
+
+    /***********************************************************************
+
+        Return the length of the internal buffer
+
+    ***********************************************************************/
+
+    size_t length () nothrow
+    {
+        this.lock.lock_nothrow();
```
AndrejMitrovic: @omerfirmak my concurrency-foo isn't the best.

Can you help me understand the purpose of using a lock when looking up a value like this? Because as soon as the function returns the lock is released, which means the state of the channel could change while the calling code still assumes the channel is in the same state.

Isn't it the same to just return the value directly?

AndrejMitrovic: Also I feel like for checking lengths and booleans we should use `atomicStore` and `atomicLoad`, because locking is expensive.

source/geod24/concurrency.d
```diff
@@ -1446,3 +1502,665 @@ private:
...
+        Create and enqueue a `ChannelQueueEntry` to the given entryq
+
+        Assumes that the Channel is locked, unlocks it before returning
+
+        Param:
+            entryq = Queue to append the new ChannelQueueEntry
+            pVal =  Pointer to the message buffer
+            sel_state = SelectState object to associate with the
+                        newly created ChannelQueueEntry
+            sel_id = id of the select call creating the new ChannelQueueEntry
```
AndrejMitrovic: Wrong docs, there are only 2 params.

source/geod24/concurrency.d
```diff
@@ -1446,3 +1502,665 @@ private:
...
+            caller_sel_id = id of the select call being executed
+
+        Return:
+            a valid ChannelQueueEntry or null
+
+    ***********************************************************************/
+
+    private ChannelQueueEntry dequeueEntry (ref DList!ChannelQueueEntry entryq) nothrow
+    {
+        while (!entryq.empty())
```
AndrejMitrovic: This should be an `if`.

source/geod24/concurrency.d
```diff
@@ -1446,3 +1502,665 @@ private:
...
+            {
+                reader_sum += readVal;
+                if (count++ == n) chn1.close();
+            }
+
+            thread_joinAll();
+            assert(reader_sum == writer_sum);
+        }
+    );
+}
```
AndrejMitrovic: I think you'll need to change your text editor settings. Needs trailing newline. :>

source/geod24/concurrency.d
```diff
@@ -1503,6 +1503,152 @@ private:
...
+/***********************************************************************
+
+    A common interface for objects that can be used in `select()`
+
+***********************************************************************/
+
+public interface Selectable
+{
+    void selectWrite (void* ptr, SelectState sel_state, int sel_id);
+    void selectRead (void* ptr, SelectState sel_state, int sel_id);
```
AndrejMitrovic: Docs please.

source/geod24/concurrency.d
```diff
@@ -1503,6 +1503,152 @@ private:
...
+    void* select_data;
+
+    this (Selectable selectable, void* select_data)
+    {
+        this.selectable = selectable;
+        this.select_data = select_data;
+    }
+}
+
+public alias SelectReturn = Tuple!(bool, "success", int, "id");
```
AndrejMitrovic: docs, because it's public

source/geod24/concurrency.d
```diff
@@ -1503,6 +1503,152 @@ private:
...
+
+public struct SelectEntry
+{
+    /// Reference to a Selectable object
+    Selectable selectable;
+
+    /// Pointer to the select data
+    void* select_data;
+
+    this (Selectable selectable, void* select_data)
```
AndrejMitrovic: docs please, otherwise it might seem like you can partially construct it via `SelectEntry(selectable)`.

source/geod24/concurrency.d
```diff
@@ -1503,6 +1503,152 @@ private:
...
+
+/***********************************************************************
+
+    Block on multiple `Channel`s
+
+    Only one operation is completed per `select()` call
+
+    Params:
+        read_list = List of `Channel`s to select for read operation
+
```
AndrejMitrovic: boop

source/geod24/concurrency.d
```diff
@@ -1503,6 +1503,152 @@ private:
...
+public SelectReturn select (ref SelectEntry[] read_list, ref SelectEntry[] write_list)
+{
+    auto ss = new SelectState();
+    int sel_id = 0;
+
+    foreach(ref entry; read_list)
+    {
+        entry.selectable.selectRead(entry.select_data, ss, sel_id++);
+    }
+
```
AndrejMitrovic, line 1839: Does `sel_id` maybe have to be reset to 0 here before the second foreach loop?

Because if so, I would suggest to just use `foreach (sel_id, ref entry; *_list)` in both foreach loops.

source/geod24/concurrency.d
```diff
@@ -1503,6 +1503,152 @@ private:
...
+    Holds the state of a group of `selectRead`/`selectWrite` calls.
+    Synchronizes peers that will consume those calls, so that only one
+    `selectRead`/`selectWrite` call is completed.
+
+***********************************************************************/
+
+final private class SelectState
+{
+
```
AndrejMitrovic: boop

source/geod24/concurrency.d
```diff
@@ -1503,6 +1503,152 @@ private:
...
+    /***********************************************************************
+
+        Tries to atomically consume a `SelectState` and sets `id` and `success`
+        fields
+
+        Param:
+            id = ID of the `selectRead`/`selectWrite` call that is consuming
+                    this `SelectState`
+
```
AndrejMitrovic: extra blank line

source/geod24/concurrency.d
```diff
@@ -1608,22 +1754,25 @@ final public class Channel (T)
 
     ***********************************************************************/
 
-    private bool tryWrite () (auto ref T val) nothrow
+    private bool tryWrite () (auto ref T val, SelectState caller_sel = null, int caller_sel_id = 0) nothrow
```
AndrejMitrovic: I see that some of the documentation for these parameters was actually introduced in the previous commit. This should be fixed.

source/geod24/concurrency.d
```diff
@@ -1632,6 +1781,42 @@ final public class Channel (T)
...
+
+        Try to write a message to Channel without blocking and if it fails,
+        create a write queue entry using the given `sel_state` and `sel_id`
+
+        Param:
+            val = Message to write to channel
+
+            sel_state = SelectState instace of the select call being executed
+
+            sel_id = id of the select call being executed
```
AndrejMitrovic: extra blank lines

source/geod24/concurrency.d
```diff
@@ -1632,6 +1781,42 @@ final public class Channel (T)
...
+        assert(ptr !is null);
+        assert(sel_state !is null);
+        T* val = cast(T*) ptr;
+
+        this.lock.lock_nothrow();
+
+        bool success = tryWrite(*val, sel_state, sel_id);
+
+        if (!sel_state.isConsumed())
+            enqueueEntry(this.writeq, val, sel_state, sel_id);
```
AndrejMitrovic: is this a global function? Otherwise maybe use `this.enqueueEntry`.

source/geod24/concurrency.d
```diff
@@ -1705,18 +1890,26 @@ final public class Channel (T)
...
+            // if dequeueEntry fails, we will try to consume caller_sel again.
+            if (!caller_sel || write_ent || caller_sel.tryConsume(caller_sel_id))
+            {
+                output = this.buffer.front();
+                this.buffer.popFront();
 
-            if (write_ent)
+                if (write_ent)
+                {
+                    this.buffer.putBack(*(write_ent.pVal));
```
AndrejMitrovic: `this.buffer.putBack(*write_ent.pVal);`

source/geod24/concurrency.d
```diff
@@ -1735,6 +1930,42 @@ final public class Channel (T)
...
+
+        Try to read a message from Channel without blocking and if it fails,
+        create a read queue entry using the given `sel_state` and `sel_id`
+
+        Param:
+            val = Buffer to write the message to
+
+            sel_state = SelectState instace of the select call being executed
+
+            sel_id = id of the select call being executed
```
AndrejMitrovic: extra blank lines

source/geod24/concurrency.d
```diff
@@ -1851,11 +2095,11 @@ final public class Channel (T)
 
     ***********************************************************************/
 
-    private ChannelQueueEntry enqueueEntry (ref DList!ChannelQueueEntry entryq, T* pVal) nothrow
+    private ChannelQueueEntry enqueueEntry (ref DList!ChannelQueueEntry entryq, T* pVal, SelectState sel_state = null, int sel_id = 0) nothrow
```
AndrejMitrovic: This line is too long now.

source/geod24/concurrency.d
```diff
@@ -1887,13 +2131,41 @@ final public class Channel (T)
 
     ***********************************************************************/
 
-    private ChannelQueueEntry dequeueEntry (ref DList!ChannelQueueEntry entryq) nothrow
+    private ChannelQueueEntry dequeueEntry (ref DList!ChannelQueueEntry entryq, SelectState caller_sel = null, int caller_sel_id = 0) nothrow
```
AndrejMitrovic: ditto line too long

source/geod24/concurrency.d
```diff
@@ -1030,6 +1023,7 @@ protected:
 
         void notify () nothrow
         {
+            this.freezeTimer();
```
AndrejMitrovic: Why do you call it `freeze` rather than `stop`? Do you ever resume that same timer? That's what it seems to imply.

source/geod24/concurrency.d
```diff
@@ -1047,15 +1041,55 @@ protected:
...
+            return cas(&this.timer_state, TimerState.Running, TimerState.Frozen);
+        }
+
+        /***********************************************************************
+
+            Query if the internal timer was timed out
+
+        ***********************************************************************/
+
+        bool wasTimedOut () nothrow
```
AndrejMitrovic: hasTimedOut?

source/geod24/LocalRest.d
```diff
@@ -110,30 +110,6 @@ import std.traits : fullyQualifiedName, Parameters, ReturnType;
 import core.thread;
 import core.time;
 
-/// Request / Response ID
```
AndrejMitrovic, line 113: I wish the moving of the symbols was in another commit. It makes it very hard to see what actually changed vs what was only moved.

AndrejMitrovic (changes requested):
Most of my comments were about atomicity of commits and some missing docs.

Actually it's a bit hard to review because there's a lot of code changes. Hopefully @Geod24 can give it a review some time too.

Overall though it looks good. You've added tests for most code which is great. Thanks a lot for the work so far!

source/geod24/concurrency.d
omerfirmak: You are right. As long as every operation (load/store) on the variables are atomic, there is no need to use a lock in these getters. We can do this easily with the closed. But since the length is managed by the ring buffer implementation, I did not want to prematurely optimize. We can definitely get rid of those once the implementation is complete, feature and correction wise.

source/geod24/concurrency.d
omerfirmak: Yep, that must have left over while manually I split the code into commits :(

source/geod24/concurrency.d
AndrejMitrovic: Good idea. Safety first.

source/geod24/concurrency.d
omerfirmak: They were initially marked as a property but I ran into a problem. Will check it again.

source/geod24/concurrency.d
omerfirmak: That should be scoped regardless, just to be safe.

source/geod24/concurrency.d
AndrejMitrovic: I wonder if it's something to do with overloads.

Well, `@property` is a bit weird because it's got some edge-cases. You might have ran into them.

source/geod24/concurrency.d
omerfirmak: With no space in between those lines the doc generator did not put each param in its own line.

AndrejMitrovic: Huh? That sounds like a bug. :/

source/geod24/concurrency.d
omerfirmak: No, `sel_id` returned is the index of the SelectEntry in the `read_list ~ write_list`. It should be documented better.

source/geod24/concurrency.d
omerfirmak: No particular reason and they dont ever get resumed. Stop would have been more appropriate. Will change.

source/geod24/concurrency.d
omerfirmak: Me englando no good :)

source/geod24/LocalRest.d
omerfirmak: I will keep that in mind.

source/geod24/concurrency.d
omerfirmak: Are there any examples of this? Cant find a special keyword for template documentation in https://dlang.org/spec/ddoc.html

omerfirmak:
Thank you for the review @AndrejMitrovic ! All of your comments should be addressed now.

source/geod24/concurrency.d
Geod24: It's not special, it is simply a parameter for the documentation generator.

source/geod24/LocalRest.d
```diff
@@ -410,6 +331,61 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
...
+    /// Data sent by the caller
+    private struct Command
+    {
+        /// Channel to be used when responding to this Command
+        RestChannel resp_chn;
+        /// Method to call
+        string method;
+        /// Serialized arguments to the method
+        SerializedData args;
+    }
```
Geod24: The reason those were outside of any template was to avoid template bloat.
In this case, the first struct (`ShutdownCommand`) can be moved here, but then, you should remove the template parameter and pick the one from the context, otherwise there will always be 2 (equal) template parameter (not bloat, but useless).

IIUC, the reason you moved `Command` in this function was to have access to `RestChannel`, and by extension, `Variant`. My instinct would be to rather move `Variant` *outside* of this function.

Now if you do so, you'll need to template `Command` on the channel type. But actually, you don't! The reason to have an `id` in the command was to allow respond, since we didn't have channels. Now that we have this, the node can just respond on the channel it received on.

Geod24: Wait I see `ID` has been removed. That's odd. `RestChannel` is the "connection", `ID` is used to ensure re-entrancy if multiple queries are run in parallel.

source/geod24/LocalRest.d
```diff
@@ -440,8 +416,12 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
...
         string file = __FILE__, int line = __LINE__)
     {
-        auto childTid = C.spawn(&spawned!(Impl), file, line, args);
-        return new RemoteAPI(childTid, timeout);
+        auto chn = new RestChannel(16);
+        new Thread(
+        {
+            spawned!(Impl)(chn, file, line, args);
+        }).start();
+        return new RemoteAPI(chn, timeout);
```
Geod24: Why is the channel buffered ?

source/geod24/LocalRest.d
```diff
@@ -532,9 +513,7 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
 
     ***************************************************************************/
 
-    private static void spawned (Implementation) (
-        C.Tid self, string file, int line, CtorParams!Implementation cargs)
-        nothrow
+    private static void spawned (Implementation) (RestChannel chn, string file, int line, CtorParams!Implementation cargs) nothrow
```
Geod24: Indentation. Also need to change the `Params` documentation. Note that I tried quite hard to keep a parallel with the way a TCP (/HTTP) server would work (listen & co) to make it easier to reason about.

source/geod24/LocalRest.d
```diff
@@ -410,6 +331,61 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
 {
     static assert (!serializerInvalidReason!(S).length, serializerInvalidReason!S);
 
+    alias RestChannel = C.Channel!Variant;
```
Geod24: I highly regret picking the name "localrest" for this repository... When I started, I just thought "Hey I'm going to mock `vibe.web.rest`"... After almost 2 years, it is very clear that it's not just mocking a REST generator, but a full network (it supports custom serialization, has a network registry, support simultaneous connections, etc...). Hence a more apt name would be "Connection".

source/geod24/LocalRest.d
```diff
@@ -551,22 +530,6 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
             }
         }
 
-        // very simple & limited variant, to keep it performant.
-        // should be replaced by a real Variant later
```
Geod24, line 555: I'm glad we never used the real `Variant` ;)

source/geod24/LocalRest.d
```diff
@@ -410,6 +331,61 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
...
+        ubyte tag;
+
+        /// Type of a request
+        enum Type
+        {
+            command,
+            response,
+            filter,
+            time_command,
+            shutdown_command
```
Geod24: Two things:
1) In our style guide, `enum` names are Snake (e.g. `Command`). Not a problem since they're namespaced.
2) I tend to use a trailing comma at the end, just to make it more change-friendly (D allows trailing commas pretty much anywhere and it's awesome).

source/geod24/LocalRest.d
```diff
@@ -410,6 +331,61 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
...
+        union
+        {
+            Command cmd;
+            Response res;
+            FilterAPI filter;
+            TimeCommand time;
+            ShutdownCommand!API shutdown;
+        }
+
+        ubyte tag;
```
Geod24: If you define an `enum`, why is this not typed ?

source/geod24/LocalRest.d
```diff
@@ -410,6 +331,61 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
...
+
+        /// Type of a request
+        enum Type
```
Geod24: `enum Type : ubyte`

source/geod24/LocalRest.d
```diff
@@ -595,61 +558,51 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
...
+                                TimeCommand s = msg.time;
+                                control.sleep_until = Clock.currTime + s.dur;
+                                control.drop = s.drop;
+                                break;
+                            case Variant.Type.filter:
+                                FilterAPI filter_api = msg.filter;
+                                control.filter = filter_api;
+                                break;
+                            default:
+                                assert(0);
```
Geod24: Note: `enum` are absolutely effing bs because you need to sanitize them EVERY. EFFING. TIME.
So even if your tag should only take values `[0 .. 4]` if correctly typed, you can actually receive `[0 .. 255]`.
`final switch` allows you to omit the `default` if you handle all values, but obviously that's just bad.
In this case, throwing an `Error` with the enum value would be good (there's a catch at the top level that will print an error message).

Geod24: Rule of thumb: Assume input to your function is provided by a vengeful ex with a PhD in cybersecurity.

source/geod24/LocalRest.d
```diff
@@ -698,17 +652,18 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
         In order to instantiate a node, see the static `spawn` function.
 
         Params:
-          tid = `geod24.concurrency.Tid` of the node.
+          listen_chn = `RestChannel` of the node.
```
Geod24: Same remark here, connection seems more appropriate

source/geod24/LocalRest.d
```diff
@@ -776,7 +734,10 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
         public void restart (void function (API) callback = null)
             @trusted
         {
-            C.send(this.childTid, ShutdownCommand!API(callback, true));
+            if (isMainThread())
+                scheduler.start({ this.listen_chn.write(Variant(ShutdownCommand!API(callback, true))); });
+            else
+                this.listen_chn.write(Variant(ShutdownCommand!API(callback, true)));
```
Geod24: I see many usages of this pattern... Why didn't you simply write a private method in the control interface which accepts a `Variant` and abstract this away ?

source/geod24/LocalRest.d
```diff
@@ -935,56 +905,33 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
...
+                        Response sendCommandAndWaitResponse ()
+                        {
+                            Variant var;
+                            if(!this.listen_chn.write(Variant(command)))
+                                throw new Exception("Connection with peer closed");
+                            if (resp_chn.read(var, this.timeout))
+                                return var.res;
+                            else
+                                return Response(Status.Timeout, SerializedData("Request timed-out"));
+                        }
```
Geod24: Ah, here it is! That's the main issue: Sending a query to the node should not create a new channel. There should be two kinds of channels:
- 1) When `spawn`ing, you get a `BindChan` (or whatever name you see fit): this is the channel the node listens to for incoming connections.
- 2) When creating a client, you get a `ConnectionChan` (or `Connection`, or whatev) that correspond to the ephemeral connection between a client and a server (what you get after the server has `accept[4]()`ed the connection).

source/geod24/LocalRest.d
```diff
@@ -2275,7 +2222,7 @@ unittest
     node.ctrl.shutdown();
     thread_joinAll();
 }
-
+/*
```
Geod24: Sneaky, but no

source/geod24/Registry.d
```diff
@@ -15,8 +15,8 @@ import geod24.concurrency;
 /// Ditto
 public shared struct Registry
 {
-    private Tid[string] tidByName;
-    private string[][Tid] namesByTid;
+    private Object[string] objectByName;
+    private string[][Object] namesByObject;
```
Geod24: I believe, if you apply the suggestion to have a channel for the "listen" operation, you can have this typed (and same for the other `Object` types in `localrest`

source/geod24/concurrency.d
```diff
@@ -1003,11 +1035,14 @@ protected:
...
         {
-            this.registerToInfoFiber();
             if (period != Duration.init)
                 this.limit = MonoTime.currTime + period;
 
-            FiberScheduler.yield();
+            if (this.shouldBlock())
+            {
+                this.registerToInfoFiber();
+                FiberScheduler.yieldAndThrow(new FiberBlockedException());
```
Geod24: A trivial optimization would be to cache this instance, right ?

source/geod24/concurrency.d
```diff
@@ -1097,59 +1133,124 @@ private:
...
+                InfoFiber cur_fiber = this.readyq.front();
+                this.readyq.removeFront();
+
+                assert(cur_fiber.state != Fiber.State.TERM);
+
+                auto t = cur_fiber.call(Fiber.Rethrow.no);
+
+                // Fibers that block on a FiberBinarySemaphore throw an
+                // exception for scheduler to catch
+                if (auto block_ex = cast(FiberBlockedException) t)
```
Geod24: If you cache the instance (assuming there's only one "in flight" at any time), you can also use `is` here, which does pointer comparison (as opposed to dynamic cast)

source/geod24/concurrency.d
```diff
@@ -1501,6 +1501,176 @@ private:
...
+            selectable = A selectable interface reference
+            select_data = pointer to the data for the select operation
+
+    ***********************************************************************/
+
+    this (Selectable selectable, void* select_data)
+    {
+        this.selectable = selectable;
+        this.select_data = select_data;
+    }
```
Geod24: D gives you a default ctor for struct. So I assume you only want it to be constructed with two parameter ?

Geod24 (changes requested):
Amazing work! Sorry it took me so long to review. I focused on the commit that adapts the library, as it's the easiest to reason about. And as mentioned, there's a few issues in the approach, but nothing that can't be fixed :)

Geod24:
Added the last commit to v0.x.x

source/geod24/concurrency.d
AndrejMitrovic: This was my thinking too. It prevents building a partially constructed aggregate. Kind of like `explicit` in C++.

source/geod24/LocalRest.d
AndrejMitrovic: And here I thought she only cooked good pasta.

source/geod24/LocalRest.d
omerfirmak: Is the style guide posted somewhere?

AndrejMitrovic: Part of it is here: https://github.com/bpfkorea/agora/blob/v0.x.x/doc/Style.md

There may be some missing notes though.

source/geod24/concurrency.d
omerfirmak: Yes exactly, both parameters are required.

source/geod24/LocalRest.d
omerfirmak: Well, my initial motivation was just to get a field test for channels. So I didn't want to re-design localrest around channels. I just replaced the obvious stuff. What you suggest is much better obviously :) I will give it a try.

source/geod24/LocalRest.d
omerfirmak: To allow multiple requests to be queued. The requester would wait on the response channel. I am not sure if that would have any performance benefits though. This probably will have to be unbuffered for the suggestion you made below to work.

source/geod24/LocalRest.d
omerfirmak: With a per command response channel, you should not need an id for re-entrancy. For the exact same behaviour, connection channel should be flushed when node restarts though.

source/geod24/LocalRest.d
omerfirmak: It must have carried over from the old code. Will fix.

source/geod24/LocalRest.d
Geod24: Per command ? You mean per type or per instance ? Assuming it's per instance, it's bringing its own host of problems.

omerfirmak: In the current implementation, each node has a channel that it listens for commands and each command comes with a channel that the node should respond on. So it is per command instance, for now.

omerfirmak: > My instinct would be to rather move Variant outside of this function.

If you move the Variant outside of the class, wouldn't you loose the API information? This, the structs being moved, was all because of the ShutdownCommand being templated.

source/geod24/LocalRest.d
omerfirmak: @Geod24 , I gave this approach a shot today and I have some doubts about it. Having a TCP/HTTP like operation using Channels might be more trouble than it's worth. After giving it some thought, the benefits of this is kinda lost on me too :smile:  To me, it seems like only benefit is getting rid of the sender Tid in `Command`. We still have to route responses to whichever Fiber is waiting for it and we now have to manage 4 channels (ie. 2 `Connection`s, since both nodes in a pair can be the server) for **every node pair** in the system, instead of 1 `MessageBox` per node.

Geod24: > and we now have to manage 4 channels

4 channels ? Are you using channels as full-duplex connections or do you actually have a read and a write ?

Some aspects must be a bit tricky to manage, for example:
1) The `accept` channel cannot be full duplex, so a node actually need to send a channel or some kind of identifier.
2) Two fibers (client/server) can race on a single connection if used as full-duplex. This is alright as long as there is a way to know who sent/must receive the message, and they are able to put it back in the queue.

> instead of 1 `MessageBox` per node.

The main benefit is to be able to properly type the messages (the previous implementation completely bypasses the type system). There are other smaller benefit (easier to reason about, perhaps ability to control contention), but IMO type-safety is the main one.

Also I still don't see how even 4 channels / node brings overhead as compared to spawning a channel per query. I expect nodes will perform way more than 4 queries during their lifetime. Additionally, the overhead of a node is much higher than what the channel will bring.

omerfirmak: > 4 channels ? Are you using channels as full-duplex connections or do you actually have a read and a write ?

For async communication, we would need 2 channels. One to send the request, second one to wait the response on. This makes a `Connection`. They get created when a new instance of `RemoteAPI` is created and `Connection` gets sent across the `accept` channel. This `Connection` is only good to send a request to the owner thread of the `RemoteAPI`. Every client should "initiate" a `Connection` to the server by instantiating a `RemoteAPI`, thus 2 `Connections` per every node pair. Using the same `Connection` for passing requests and responses in both directions would cause Nodes to race for initiating the `Connection` and require a mechanism to keep track of already existing `Connections`, but I am not sure how big of a problem that might be. This is almost a complete network stack :smile: 

> 4 channels / node brings

It is actually 4 channels / every possible node pair. 

Overhead of having persistent `Connection`s instead of short living Channels per query is in the implementation complexity. Node have to both manage incoming and outgoing `Connection`s separately. IMO, the biggest drawback of spawning a Channel per query is creating the OS Mutex and the GC allocation of the Channel itself, which I am not sure if either would be a significant bottleneck. But, since we know the lifetime of the response Channel we can allocate it on the stack. Also, OS Mutex can be replaced with a FiberMutex easily. **EDIT:** Maybe not so easily :P

Essentially what I am trying to say is that I would prefer simplicity over potential performance increase that we probably dont need right now. I might have gotten the whole `Connection` design wrong tho, it might be possible to implement it way simpler than I am trying to right now. :smile:

source/geod24/LocalRest.d
Geod24: > If you move the Variant outside of the class, wouldn't you loose the API information? This, the structs being moved, was all because of the ShutdownCommand being templated.

Well you can do `Variant(API)` and have an alias here.

source/geod24/LocalRest.d
Geod24: Unbuffered also means the sender will not block, won't that create issues ?

omerfirmak: Do you mean buffered? Currently it wont. the `read` on the response Channel will block the sender.

omerfirmak:
~~I have just added the code for FiberMutex and FiberSemaphore but I am getting weird deadlocks in Agora tests. Do you guys see anything obviously wrong?
with OS locks, Fibers of same thread get to lock a Mutex at the same time, with FiberMutex they wont be able to do that. The bug might not be in the FiberMutex itself.~~

source/geod24/LocalRest.d
Geod24: > Using the same `Connection` for passing requests and responses in both directions would cause Nodes to race for initiating the `Connection` and require a mechanism to keep track of already existing `Connections`, but I am not sure how big of a problem that might be.

I'm not sure I follow. So let's say we have a network with 2 nodes, `A` and `B`. 
A` connects as a client to `B`, and `B` connects as a client to `A`.
`A` now have two `Connection`s: the client connection to `B`, and the server `Connection` to `B`. So does `B`.

When nothing happens, nodes listen to their server connection. When they get a query, they process it, and then write back on that same connection.
When doing a request as a client (e.g. during the processing of a query), they first write to the client connection. This should immediately block (yield to the event loop) until the query has been read. Then, they  read from that channel, which blocks (yield to the event loop) until someone writes.

The only moment nodes could ever race is when there's recursion.
E.g. `A [client]` -> `[server] B [client]` -> `[server] A [client]` -> `[server] B`.

However, you can match the  waiting fiber with the response if you associate them with an ID.

omerfirmak: Moar commits for you to review @Geod24 @AndrejMitrovic  :)
The last commit should eventually replace [this](https://github.com/Geod24/localrest/pull/95/commits/91492cd1682efca7caec0fe8e9b72aa2960de6c9).

source/geod24/concurrency.d
```diff
@@ -963,6 +963,44 @@ class FiberScheduler
...
+        assert(info_fiber, "Called from outside of an InfoFiber");
+        info_fiber.resources.insert(r);
+    }
+
+    /***********************************************************************
+
+        Remove Resource from the Resource list of runnning Fiber
+
+        Param:
+            r = Resource instace
```
AndrejMitrovic: What does it return?

source/geod24/concurrency.d
```diff
@@ -1242,6 +1288,27 @@ private:
         return earliest_timeout;
     }
 
+    /***********************************************************************
+
+        Release all resources currently held by Fibers
```
AndrejMitrovic: by Fibers => by the provided running fiber?

source/geod24/concurrency.d
```diff
@@ -2230,6 +2316,19 @@ final public class Channel (T) : Selectable
...
+        /***********************************************************************
+
+            Terminate ChannelQueueEntry so that it is neutralized
+
+        ***********************************************************************/
+
+        void release() nothrow
+        {
+            if (this.blocker.stopTimer())
+                this.pVal = cast(T*) 0x0; // Sanitize pVal so that we can catch illegal accesses
```
AndrejMitrovic: I think this can just be `= null;`?

AndrejMitrovic: The cast will make the code non-`@safe`. It would be good to make most of the code we write `@safe`. But we can do it later too..

source/geod24/LocalRest.d
```diff
@@ -148,13 +148,227 @@ private enum Status
...
+        Params:
+            res = Newly arrived Response
+
+    *******************************************************************************/
+
+    void notifyWaiter (Response res)
+    {
+        // Is the waiting Fiber already blocked?
+        auto waiter = (res.id in this.waiting_list);
+        if (waiter !is null)
```
AndrejMitrovic: trick: 

```D
if (auto waiter = res.id in this.waiting_list)
{
}
else ...
```

source/geod24/LocalRest.d
```diff
@@ -148,13 +148,227 @@ private enum Status
...
+            waiter.res = res;
+            waiter.blocker.notify();
+        }
+        else // Fiber is not yet blocked, create an entry for the related Fiber to use
+            this.waiting_list[res.id] = Waiting(null, res);
+    }
+
+    /*******************************************************************************
+
+        Close the `Channel`s associated with this `Connection`
```
AndrejMitrovic: It would be good to document what happens after the channel is closed. Do the other nodes's requests fail, or time out?

source/geod24/LocalRest.d
```diff
@@ -443,10 +602,11 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
             cmd    = the command to run (contains the method name and the arguments)
             node   = the node to invoke the method on
             filter = used for filtering API calls (returns default response)
+            resp_chn = `Channel` to send the `Response` from
```
AndrejMitrovic: send to? or read from?

send from doesn't sound right.

source/geod24/LocalRest.d
```diff
@@ -553,31 +711,78 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
             // Control the node behavior
             Control control;
 
+            struct AwaitingMessages
```
AndrejMitrovic: AwaitingMessages => AwaitingMessage

It's only a single message, and `await_msgs` is the array.

source/geod24/LocalRest.d
```diff
@@ -553,31 +711,78 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
...
-                while (1)
+                while (true)
                 {
+                    Connection new_conn;
+                    Response res;
                     Variant msg;
-                    if (chn.read(msg, 10.msecs))
+
+                    C.SelectEntry[] read_list;
+                    C.SelectEntry[] write_list;
```
AndrejMitrovic: This will be quite GC-heavy with all the allocations. How about moving it oustide the while loop, resetting the lengths to zero and using `assumeSafeAppend`?

source/geod24/LocalRest.d
```diff
@@ -553,31 +711,78 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
...
+                            handleCommand(await_msg.var.cmd, node, control.filter, await_msg.conn.resp_chn);
+                        await_msgs.length = 0;
+                        assumeSafeAppend(await_msgs);
+                    }
+
+                    if (!sel_ret.success)
+                        continue;
+
+                    // Bind chn
+                    if (auto bind_chn = cast(BindChn) read_list[sel_ret.id].selectable)
```
AndrejMitrovic: `bid_chn` doesn't seem to be used here. So how about changing it to `if (cast(BindChn) read_list[sel_ret.id].selectable)`?

source/geod24/LocalRest.d
```diff
@@ -553,31 +711,78 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
...
+                    {
+                        incoming_conns[new_conn.command_chn] = new_conn;
+                    }
+                    // Command
+                    else if (auto comm_chn = cast(CommandChn) read_list[sel_ret.id].selectable)
                     {
+                        auto curr_conn = incoming_conns[comm_chn];
```
AndrejMitrovic, line 768: This can never be null?

omerfirmak: Yep, that is non-null by design.

omerfirmak:
The weird thing is; with the last Connection changes, I am getting a 2x peak memory usage. The heap usage is barely increased. I think that the last changes are somehow holding the Fibers from getting GC'ed.

AndrejMitrovic:
> The weird thing is; with the last Connection changes, I am getting a 2x peak memory usage. The heap usage is barely increased. I think that the last changes are somehow holding the Fibers from getting GC'ed.

Do you know about the `-vgc` flag? It will list all allocations in the code.

source/geod24/LocalRest.d
```diff
@@ -586,73 +706,110 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
         void runNode ()
         {
             scope node = new Implementation(cargs);
-            scheduler = new LocalScheduler;
+            scheduler = new C.FiberScheduler;
 
             // Control the node behavior
             Control control;
 
+            struct AwaitingMessage
```
AndrejMitrovic, line 713:
```suggestion
            static struct AwaitingMessage
```

D likes to add the context pointer even if it's unnecessary.

source/geod24/LocalRest.d
```diff
@@ -1490,6 +1610,9 @@ unittest
 
         override void callFoo()
         {
+            if (!this.remote)
```
AndrejMitrovic, line 1612: Why has this changed? It doesn't work when it's in the constructor?

omerfirmak: it doesnt, `node_chn` gets initialized only after the first `spawn` finishes. Constructor runs before that.

source/geod24/concurrency.d
```diff
@@ -926,84 +1038,138 @@ protected:
...
 
-        bool wait(Duration period) nothrow
+        ***********************************************************************/
+
+        bool wait (Duration period = Duration.init) nothrow
         {
-            this.registerToInfoFiber();
-            this.limit = MonoTime.currTime + period;
-            FiberScheduler.yield();
+            if (period != Duration.init)
```
AndrejMitrovic, line 1091: What about `Duration.max`?

source/geod24/LocalRest.d
```diff
@@ -182,105 +148,268 @@ private enum Status
...
 {
-    static if (T.length == 0)
-        size_t dummy;
-    T args;
+    /// Sequence id of Command
+    size_t id;
+    /// Method to call
+    string method;
+    /// Serialized arguments to the method
+    SerializedData args;
```
Geod24: Can you move them back to where  they were so they don't show as diff ?

source/geod24/concurrency.d
AndrejMitrovic: I mean, I think I saw code which handled both cases.

source/geod24/concurrency.d
```diff
@@ -1012,56 +1178,154 @@ private:
...
+
+        MonoTime earliest_timeout;
 
-        assert(pos < m_fibers.length);
-        while (m_fibers.length > 0)
+        while (!this.readyq.empty() || !this.wait_list.empty())
         {
-            // Is Fiber waiting on a FiberBinarySemaphore?
-            if (auto sem = m_fibers[pos].sem)
+            while (!readyq.empty())
```
AndrejMitrovic, line 1200: I'm a little bit confused about the nested loop. Can't you just remove it? The outer while loop handles everything.

source/geod24/concurrency.d
```diff
@@ -1012,56 +1178,154 @@ private:
...
+                if (MonoTime.currTime >= earliest_timeout || this.sem.tryWait())
+                    earliest_timeout = wakeFibers();
             }
 
-            auto t = m_fibers[pos].call(Fiber.Rethrow.no);
-            if (t !is null)
+            if (!this.wait_list.empty())
             {
-                throw t;
+                Duration time_to_timeout = earliest_timeout - MonoTime.currTime;
```
AndrejMitrovic, line 1258: Is it guaranteed that `earliest_timeout` will not equal `Duration.init` here? Actually I'm not even sure what happens when we subtract against that.

source/geod24/concurrency.d
```diff
@@ -1446,3 +1710,1545 @@ private:
...
+
+            entryq.removeFront();
+        }
+
+        return null;
+    }
+
+private:
+    /// Internal data storage
+    FixedRingBuffer!T buffer;
```
Geod24: Do we really need to bring in a 315 LoC file just for this ? `FixedRingBuffer` seems rather generic, but we only iterate a single item at once and never resize the buffer. I only see 4 methods being used, `putBack`, `popFront`, `length`, `front`, and the initial capacity set.

source/geod24/concurrency.d
```diff
@@ -1012,56 +1178,154 @@ private:
...
+
+            // Remove the unblocked Fiber from wait list and
+            // append it to the end ofready queue
+            if (!fiber.blocker.shouldBlock())
             {
-                if (pos >= (m_fibers = remove(m_fibers, pos)).length)
-                    pos = 0;
+                this.wait_list.popFirstOf(wait_range);
+                fiber.blocker = null;
+                this.readyq.insert(fiber);
```
AndrejMitrovic, line 1297: What about popping from the `wait_range`? It's not needed here?

source/geod24/concurrency.d
omerfirmak: Yes, `earliest_timeout` is always > 0

source/geod24/concurrency.d
omerfirmak: `popFirstOf` does that too. It removes both from the list and the range.

source/geod24/concurrency.d
omerfirmak: Only `Duration.init` is considered infinite timeout throughout localrest and concurrency.d. You must have seen it elsewhere?

source/geod24/LocalRest.d
omerfirmak: ~~Once we squash the two LocalRest.d adaptation commits, they should be gone. I wanted to keep the two versions in the history.~~ Oh I see what you mean now :smile:

source/geod24/concurrency.d
omerfirmak: We would need to replace the inner `while` with `if` and add an additional check to the `if` below to avoid sleeping while ready queue is not empty. I guess this could work

```diff
diff --git a/source/geod24/concurrency.d b/source/geod24/concurrency.d
index 947835d..909ef15 100644
--- a/source/geod24/concurrency.d
+++ b/source/geod24/concurrency.d
@@ -1190,7 +1190,7 @@ private:
 
         while (!this.readyq.empty() || !this.wait_list.empty())
         {
-            while (!readyq.empty())
+            if (!readyq.empty())
             {
                 InfoFiber cur_fiber = this.readyq.front();
                 this.readyq.removeFront();
@@ -1233,8 +1233,7 @@ private:
                 if (MonoTime.currTime >= earliest_timeout || this.sem.tryWait())
                     earliest_timeout = wakeFibers();
             }
-
-            if (!this.wait_list.empty())
+            else if (!this.wait_list.empty())
             {
                 Duration time_to_timeout = earliest_timeout - MonoTime.currTime;
 
@@ -1243,10 +1242,10 @@ private:
                     this.sem.wait();
                 else if (time_to_timeout > 0.seconds)
                     this.sem.wait(time_to_timeout);
-            }
 
-            // OS Thread woke up populate ready queue
-            earliest_timeout = wakeFibers();
+                // OS Thread woke up populate ready queue
+                earliest_timeout = wakeFibers();
+            }
         }
```

omerfirmak:
Did a rebase. Good news, we have merged more than half of the code yesterday.

AndrejMitrovic:
> Did a rebase. Good news, we have merged more than half of the code yesterday.

Do you want to continue splitting this up into multiple PRs? I don't think I can review 20 commits in one go.

PR #98, opened 2020-11-11
Implement a RingBuffer

omerfirmak:
Using RingBuffer in Channels would decrease the
amount of GC allocations

### DIFF
--- a/source/geod24/RingBuffer.d
+++ b/source/geod24/RingBuffer.d
@@ -1,0 +1,135 @@
+/*******************************************************************************
+
+    Contains a ring buffer implementation with fixed size
+
+    Copyright:
+        Copyright (c) 2020 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module geod24.RingBuffer;
+
+import std.range;
+
+/*******************************************************************************
+
+    A ring buffer with fixed size
+
+    Params:
+        T = Type of elements
+
+*******************************************************************************/
+
+public class RingBuffer (T)
+{
+    /*******************************************************************************
+
+        Params:
+            size = Maximum number of elements, must be greater than 0
+
+    *******************************************************************************/
+
+    this (size_t size) @safe pure nothrow
+    {
+        assert(size, "Cannot create a RingBuffer with size of 0");
+        this.storage.length = size;
+        this.write_range = cycle(this.storage[]);
+        this.read_range = cycle(this.storage[]);
+    }
+
+    /// Returns: true if full
+    public bool full () @safe @nogc pure nothrow
+    {
+        return this.len == this.storage.length;
+    }
+
+    /// Returns: true if empty
+    public bool empty () @safe @nogc pure nothrow
+    {
+        return this.len == 0;
+    }
+
+    /// Returns: Number of elements currently in the buffer
+    public size_t length () @safe @nogc pure nothrow
+    {
+        return this.len;
+    }
+
+    /*******************************************************************************
+
+        Insert an element to the buffer
+
+        Params:
+            val = an element of type T
+
+    *******************************************************************************/
+
+    public void insert () (auto ref T val) @safe @nogc pure nothrow
+    {
+        assert(!this.full(), "Attempting to insert to a full RingBuffer");
+
+        write_range.front() = val;
+        write_range.popFront();
+        this.len++;
+    }
+
+    /// Returns: Element in the front of the buffer
+    public T front () @safe @nogc pure nothrow
+    {
+        assert(!this.empty(), "Attempting to read from a full RingBuffer");
+        return read_range.front();
+    }
+
+    /// Removes the element in the front of the buffer
+    public void popFront () @safe @nogc pure nothrow
+    {
+        assert(!this.empty(), "Attempting to pop from a full RingBuffer");
+        this.len--;
+        read_range.popFront();
+    }
+
+    /// Current number of elements
+    private size_t len;
+
+    /// Underlying storage
+    private T[] storage;
+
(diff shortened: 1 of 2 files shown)


source/geod24/RingBuffer.d
```diff
@@ -0,0 +1,127 @@
+module geod24.RingBuffer;
```
AndrejMitrovic, line 14: Needs copyright header.

source/geod24/RingBuffer.d
```diff
@@ -0,0 +1,127 @@
...
+
+/*******************************************************************************
+
+    A ring buffer implementation with fixed size
+
+    Params:
+        T = Type of elements
+
+*******************************************************************************/
+
```
AndrejMitrovic, line 26: Please add access specifiers (public / protected) for all symbols.

source/geod24/RingBuffer.d
```diff
@@ -0,0 +1,127 @@
...
+        Params:
+            size = Maximum number of elements
+
+    *******************************************************************************/
+
+    this (size_t size) @safe pure nothrow
+    {
+        this.storage = new T[size];
+        this.storage.length = size;
```
AndrejMitrovic: This is is sufficient:

```suggestion
        this.storage.length = size;
```

Slices don't need to be `new`-ed, although they can. But either one of these lines you wrote would do the exact same thing.

AndrejMitrovic:
Is there really no ring buffer implementation in Phobos? There wasn't the last time I looked (but that was .. 10 years ago?).

source/geod24/RingBuffer.d
```diff
@@ -0,0 +1,127 @@
...
+        if (size)
```
AndrejMitrovic: Why are we supporting ring buffers with zero capacity if using it will result in an error (exceptions thrown)?

Wouldn't it be better to just assert that size != 0?

source/geod24/RingBuffer.d
```diff
@@ -0,0 +1,127 @@
...
+    {
+        assert(!this.full());
+
+        write_range.front() = val;
+        write_range.popFront();
+        this.len++;
+    }
+
+    /// Returns: Element in the front of the buffer
+    T popFront () @safe @nogc pure nothrow
```
AndrejMitrovic: This is breaking with the known definition of a range. `popFront` should be `void`. Only `front` / `back` / `opIndex` should ever return any element. It's very confusing to see client code receive something from `popFront` - it breaks assumptions for most D programmers.

If you wish to have a function which calls both `front()` and `popFront()` then I would recommend calling it something else. I wish I could come up with a good name for it.

source/geod24/RingBuffer.d
```diff
@@ -0,0 +1,127 @@
...
+    A ring buffer implementation with fixed size
+
+    Params:
+        T = Type of elements
+
+*******************************************************************************/
+
+class RingBuffer (T)
+{
+
```
AndrejMitrovic: Blank line

omerfirmak:
> Is there really no ring buffer implementation in Phobos? There wasn't the last time I looked (but that was .. 10 years ago?).

Sadly, no. Only one that I can find is the one in vibe.d

source/geod24/RingBuffer.d
```diff
@@ -0,0 +1,127 @@
...
+    {
+        return this.len;
+    }
+
+    /*******************************************************************************
+
+        Insert an element to the buffer
+
+        Params:
+            val = new element of type T
```
AndrejMitrovic:
```suggestion
            val = an element of type T
```

source/geod24/RingBuffer.d
```diff
@@ -0,0 +1,127 @@
...
+
+// A buffer with 0 size should be both empty and full
+unittest
+{
+    auto buffer = new RingBuffer!int(0);
+    assert(buffer.full());
+    assert(buffer.empty());
+}
+
+unittest
```
AndrejMitrovic, line 108: Does it also work when you add elements, remove, and then add again? Specifically I'm wondering how the `read_range`  behaves.

source/geod24/RingBuffer.d
omerfirmak: When I first saw that `popFront` does not return anything I was disappointed :smile: Removing and reading a value from a container should be atomic IMO.

source/geod24/RingBuffer.d
omerfirmak: Just so that Channel can get away with not checking if it is operating in buffered or unbuffered mode. I will make Channel explicitly check its op mode and assert size != 0 here.

source/geod24/RingBuffer.d
omerfirmak: Added a loop to test that case

omerfirmak:
Thanks @AndrejMitrovic , all comments addressed.

source/geod24/RingBuffer.d
AndrejMitrovic: I understand. There's many people who think so. :) I think there is some history around it, but I'd have to dig up in the archives to find why it's designed this way.

source/geod24/RingBuffer.d
```diff
@@ -0,0 +1,135 @@
...
+/*******************************************************************************
+
+    A ring buffer with fixed size
+
+    Params:
+        T = Type of elements
+
+*******************************************************************************/
+
+class RingBuffer (T)
```
AndrejMitrovic: public please. :) And then feel free to merge!

AndrejMitrovic (approved):

source/geod24/RingBuffer.d
omerfirmak: Done. But, I dont have write access here.

PR #99, opened 2020-11-13
Port LocalRest to use Channels

omerfirmak:
This is the next step to fully merge #95

### DIFF
--- a/source/geod24/LocalRest.d
+++ b/source/geod24/LocalRest.d
@@ -110,24 +110,11 @@ import std.traits : fullyQualifiedName, Parameters, ReturnType;
 import core.thread;
 import core.time;
 
-/// Request / Response ID
-private struct ID
-{
-    /// A node may restart, in which case it will spawn a new request scheduler
-    size_t sched_id;
-    /// In order to support re-entrancy, every request contains an id
-    /// which should be copied in the `Response`
-    /// Initialized to `size_t.max` so not setting it crashes the program
-    size_t id = size_t.max;
-}
-
 /// Data sent by the caller
 private struct Command
 {
-    /// Tid of the sender thread (cannot be JSON serialized)
-    C.Tid sender;
-    /// ID used for request re-entrancy (See ID definition)
-    ID id;
+    /// Sequence id of Command
+    size_t id;
     /// Method to call
     string method;
     /// Serialized arguments to the method
@@ -144,10 +131,10 @@ private struct TimeCommand
 }
 
 /// Ask the node to shut down
-private struct ShutdownCommand (API)
+private struct ShutdownCommand
 {
     /// Any callback to call before the Node's destructor is called
-    void function (API) callback;
+    void function (Object) callback;
 
     /// Whether we're restarting or really shutting down
     bool restart;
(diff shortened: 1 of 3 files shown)


source/geod24/LocalRest.d
```diff
@@ -110,24 +110,11 @@ import std.traits : fullyQualifiedName, Parameters, ReturnType;
 import core.thread;
 import core.time;
 
-/// Request / Response ID
-private struct ID
-{
-    /// A node may restart, in which case it will spawn a new request scheduler
-    size_t sched_id;
```
Geod24, line 117: This is suspicious. I don't see why the restart logic should be affected here.

source/geod24/LocalRest.d
```diff
@@ -144,10 +131,10 @@ private struct TimeCommand
 }
 
 /// Ask the node to shut down
-private struct ShutdownCommand (API)
+private struct ShutdownCommand
 {
     /// Any callback to call before the Node's destructor is called
-    void function (API) callback;
+    void function (Object) callback;
```
Geod24, line 137: Why the type erasure ? We could just template `Variant`, couldn't we ?

source/geod24/LocalRest.d
```diff
@@ -212,75 +199,216 @@ private struct ArgWrapper (T...)
...
-    /// The 'Response' we are currently processing, if any
-    private Response pending;
+    Type tag;
 
-    /// Request IDs waiting for a response
-    private Waiting[ID] waiting;
+    /// Type of a request
+    enum Type : ubyte
+    {
+        command,
```
Geod24, line 226: CamelCase is the D style IIRC

source/geod24/LocalRest.d
```diff
@@ -212,75 +199,216 @@ private struct ArgWrapper (T...)
...
+private class Connection
+{
+    ///
+    this () @trusted nothrow
     {
-        static size_t last_idx;
-        this.sched_id = last_idx++;
+        this.command_chn = new CommandChn();
+        this.resp_chn = new RespChn();
     }
```
Geod24: I don't see any reason this should be `@trusted`. Either the ctors are `@safe` / `@trusted`, or this is `@system`.
`@trusted` would only make sense if we were passing parameters, but here we're not.

source/geod24/LocalRest.d
omerfirmak: Command ids persist through restarts, they are not restarted back from 0. So sched_id is not needed.

source/geod24/LocalRest.d
omerfirmak: We would need to template `CommandChn`, `Connection` and `BindChn` too, right? I wanted to avoid that.

source/geod24/LocalRest.d
omerfirmak: Yeah you are right, I might have been too generous with my `@trusted` usage.

omerfirmak:
Do we know how many hardware threads we get in Travis?

source/geod24/LocalRest.d
omerfirmak: https://dlang.org/dstyle.html says enums are camelCase

AndrejMitrovic:
> Do we know how many hardware threads we get in Travis?

There is `totalCPUs` in `std.paralellism` you can inspect. So you can verify it programmatically.

omerfirmak:
> > Do we know how many hardware threads we get in Travis?
> 
> There is `totalCPUs` in `std.paralellism` you can inspect. So you can verify it programmatically.

Thanks! Looks like we only get 2. With all the busy loops Channels are causing, performance issues are making the tests fail. We would need to merge the scheduler too, I guess.

PR #106, opened 2021-01-12
Small refactorings to help with Channels

Geod24:
Some PRs have been waiting for a while and it's about time we bring them in.


omerfirmak:
These all look good to me. They look like a subset of #99, am I right?

Geod24:
> They look like a subset of #99, am I right?

Indeed.

PR #107, opened 2021-01-12
Strongly type and differentiate the listening connection to the remote node

Geod24:
```
Since our framework follow the same model as the basic TCP/IP server,
we have two kinds of connections: the "main" connection, or listener,
which in C would be the file descriptor given by `bind(2)`.
Whenever a client wants to connect to us, they reach us through that
connection, and we establish a new client connection.
In C, that would be the file descriptor returned by `accept`,
in our framework it is a `RemoteAPI`.
```

@omerfirmak : Hope the changes here make sense. I did adapt your PR to be fully typed, in light of those, but it still times out.

### DIFF
--- a/source/geod24/LocalRest.d
+++ b/source/geod24/LocalRest.d
@@ -409,6 +409,25 @@ public final class Timer
 
 /*******************************************************************************
 
+    A reference to the "listening" connection of a remote thread
+
+    When a remote thread starts, it initially listens for new connection
+    (similarly to `bind` in C). When a new connection is started, it creates
+    a separate channel for communication (similar to `accept` in C).
+    This newly-created channel is what `RemoteAPI` wraps.
+    The channel / link / connection to the original listener, which is the only
+    one able to establish a connection, is what this data structure wraps.
+
+*******************************************************************************/
+
+public struct Listener (API)
+{
+    /// Internal data, do not use
+    package C.Tid data;
+}
+
+/*******************************************************************************
+
     A reference to an alread-instantiated node
 
     This class serves the same purpose as a `RestInterfaceClient`:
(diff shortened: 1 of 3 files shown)


source/geod24/LocalRest.d
```diff
@@ -708,16 +727,16 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
         In order to instantiate a node, see the static `spawn` function.
 
         Params:
-          tid = `geod24.concurrency.Tid` of the node.
-                This can usually be obtained by `geod24.concurrency.locate`.
+          listener = The listener used to connect to the node (most frequently
+                     obtained by ccalling `geod24.concurrency.locate`)
```
omerfirmak:
```suggestion
                     obtained by calling `geod24.concurrency.locate`)
```

source/geod24/Registry.d
```diff
@@ -15,19 +15,16 @@ module geod24.Registry;
 
 import core.sync.mutex;
 import geod24.concurrency;
+import geod24.LocalRest;
 
 /// Ditto
 public shared struct Registry (API)
```
omerfirmak, line 21: So, do we now indent to have multiple registries in a heterogeneous network? Like we would have both `Validator`s and `FullNode`s in the network.

omerfirmak (approved):

source/geod24/Registry.d
Geod24: Well, we already kind of did: We can have a registry with all nodes as `FullNode`, or just the validators as `Validator`.

omerfirmak:
Oh I didnt realize the failure in the CI. I'll take a look

omerfirmak:
It is probably because the timeouts are too low for the macos runners. I'll increase them and see if it helps.

PR #110, opened 2021-02-09
Use atomic operations to read/write closed flag

Geod24:
Some places read it without synchronization, but they are all using the mutex.
So even if we close it while another thread is reading,
nothing will be terminated until the lock is released.

### DIFF
--- a/source/geod24/concurrency.d
+++ b/source/geod24/concurrency.d
@@ -2008,7 +2008,7 @@ final public class Channel (T) : Selectable
 
         bool success = tryWrite(val);
 
-        if (!success && !this.closed)
+        if (!success && !this.isClosed())
         {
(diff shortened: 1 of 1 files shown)


source/geod24/concurrency.d
```diff
@@ -2261,10 +2261,10 @@ final public class Channel (T) : Selectable
 
     void close () nothrow
     {
-        this.lock.lock_nothrow();
```
omerfirmak, line 2264: `close()` should probably still be synchronized with a mutex, because it clears the internal queues.

omerfirmak (changes requested):

source/geod24/concurrency.d
Geod24: I just moved it inside the `if` ?

Geod24:
Updated

omerfirmak (approved):

PR #111, opened 2021-02-09
Cleanup unused parts of concurrency.d

omerfirmak:
Since LocalRest now uses Channels, most of the old code is not used anymore.


source/geod24/concurrency.d
```diff
@@ -93,7 +63,7 @@ private
         bool doesIt = false;
         static foreach (T; Types)
         {
-            static if (is(T == Tid))
+            static if (is(T : Channel!R, R))
```
Geod24: Shouldn't we check `R` for local aliasing here ?

Geod24 (approved):
Me likey, just rebase

source/geod24/concurrency.d
omerfirmak: Done

omerfirmak:
Rebased

source/geod24/concurrency.d
```diff
@@ -107,8 +77,8 @@ private
 
     @safe unittest
     {
-        static struct Container { Tid t; }
-        static assert(!hasLocalAliasing!(Tid, Container, int));
+        static struct Container { Channel!int t; }
+        static assert(!hasLocalAliasing!(Channel!int, Container, int));
```
Geod24:
```suggestion
        static assert(!hasLocalAliasing!(Channel!(Channel!int), Channel!int, Container, int));
        static assert( hasLocalAliasing!(Channel!(Channel!(int[]))));        
```

For testing the new code path ?

omerfirmak: and it breaks immediately:)

Geod24: We're passing locally aliased types around ?

omerfirmak: `hasLocalAliasing!(Channel!(Channel!int))`

It didnt like the nested channels:)

```source/geod24/concurrency.d(66,24): Error: declaration R is already defined```

Geod24: Hum, try replacing `R` with something else... E.g. `ChannelType` ?

Geod24: https://issues.dlang.org/show_bug.cgi?id=20018

Geod24:
Let's see how this fare

PR #113, opened 2021-02-25
Minor style fixes / refactoring

Geod24:
See each commit.

### DIFF
--- a/source/geod24/LocalRest.d
+++ b/source/geod24/LocalRest.d
@@ -1221,7 +1221,6 @@ unittest
     assert(test.pubkey() == 42);
     test.ctrl.shutdown(&onDestroy);
     thread_joinAll();
-    // ctr.shutdown call is asynchronous
     assert(dtor_called);
     assert(onDestroy_called);
     assert(shutdown_called);
--- a/source/geod24/concurrency.d
+++ b/source/geod24/concurrency.d
@@ -141,7 +141,7 @@ public void thisScheduler (FiberScheduler value) nothrow
  * This is an example scheduler that creates a new Fiber per call to spawn
  * and multiplexes the execution of all fibers within the main thread.
  */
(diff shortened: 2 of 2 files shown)


source/geod24/LocalRest.d
```diff
@@ -303,10 +303,12 @@ private class Connection
...
-                while (this.resp_chn.read(res, timeout))
-                    if (res.id == resp_id)
-                        return;
-                res = Response(Status.Timeout, resp_id, SerializedData("Request timed-out"));
+                if (!this.resp_chn.read(res, timeout))
+                    res = Response(Status.Timeout, resp_id, SerializedData("Request timed-out"));
+                // Otherwise the `res` variable would have been filled by `read`,
+                // and the response ID should match since the main thread never
+                // send concurrent requests
+                assert(res.id == resp_id);
```
omerfirmak: Could we hit a case similar to what we encountered in `vibe.d`?
A response for a timed out request could be waiting in the channel. Probably will never happen in practice.

source/geod24/concurrency.d
```diff
@@ -463,6 +469,11 @@ public class FiberScheduler
         thisScheduler(this);
         scope (exit) thisScheduler(null);
 
+        /// Allow recursively calling dispatch
```
omerfirmak: What is the use case for this?

omerfirmak (approved):

source/geod24/concurrency.d
Geod24: It was to allow an improvement for the scheduler in the main thread, but thinking about this more, I will take this commit out.

source/geod24/LocalRest.d
Geod24: Hum that's actually a good point, I'll look at the code once more, and will take this commit out in the meantime.

PR #115, opened 2021-03-05
Templatize Channel to enable use of different schedulers

omerfirmak:
This can enable Channels to be used with vibe.d event loop without select
support. Templatizing select causes a bunch of template mess, which I didn't want
to go through at this point.

Using select will result in a runtime crash.

### DIFF
--- a/source/geod24/concurrency.d
+++ b/source/geod24/concurrency.d
@@ -852,7 +852,8 @@ private:
 
 ***********************************************************************/
 
-final public class Channel (T) : Selectable
+final public class Channel (T, ChannelMutex = FiberMutex,
+    ChannelBlocker = FiberScheduler.FiberBlocker) : Selectable
 {
     /***********************************************************************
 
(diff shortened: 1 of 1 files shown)


omerfirmak:
nvm. this wont work. Vibe.d primitives does not provide most of the functionality we need.

omerfirmak:
Wait, they do.

AndrejMitrovic (approved):
just needs param docs (maybe mention vibe.d support)

omerfirmak:
I'll give this a try today and merge if it works as expected.

omerfirmak:
This wont work without some wrappers around vibe.d primitives.

Mostly https://github.com/vibe-d/vibe-core/blob/0c1ea531b8495b754364f145acc2b612a4598864/source/vibe/core/sync.d#L984
and https://github.com/vibe-d/vibe-core/blob/0c1ea531b8495b754364f145acc2b612a4598864/source/vibe/core/sync.d#L326

Marking this as a draft until we actually need a `VibeChannel`

PR #116, opened 2021-03-11
Allow spawning tasks in the constructor of a Node

AndrejMitrovic:
This allows a Node to spawn tasks and timers within its constructor.

This is a common problem we had in Agora that we always worked around with `run()` routines.

@omerfirmak I can't really re-use `isMainThread` here because it explicitly checks for `C.thisScheduler() is null`, which will still be `null` until `dispatch()` is called.

And I cannot change `isMainThread` either because other code depends on it having this kind of behavior.

### DIFF
--- a/source/geod24/LocalRest.d
+++ b/source/geod24/LocalRest.d
@@ -723,8 +723,9 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
 
         void runNode ()
         {
-            scope node = new Implementation(cargs);
             scheduler = new C.FiberScheduler;
+            C.thisScheduler(scheduler);
+            scope node = new Implementation(cargs);
 
             // Control the node behavior
             Control control;
@@ -2502,4 +2503,40 @@ unittest
         Listener!APIExtended(node.ctrl.listener().data));
     assert(extnode.required() == 42);
     assertThrown!ClientException(extnode.optional());
+}
+
+/// test that runTask works in the constructor
+unittest
+{
+    __gshared bool called;
+    static interface API
+    {
+    @safe:
+        public @property ulong pubkey ();
+    }
+
+    static class MockAPI : API
+    {
+    @safe:
+        this () @trusted
+        {
+            runTask(&callMe);
+        }
+
+        void callMe () @trusted
+        {
+            called = true;
+        }
+
+        public override @property ulong pubkey () { return 42; }
+    }
+
+    scope test = RemoteAPI!API.spawn!MockAPI();
+    scope (exit)
+    {
+        test.ctrl.shutdown();
+        thread_joinAll();
+    }
+    assert(test.pubkey() == 42);
+    assert(called);
 }


omerfirmak:
I think we can make an exception and set `thisScheduler` as soon as we create the first `FiberScheduler` in `runNode`. Would not that be simpler? I cant immediately think of any side affects it would have.

AndrejMitrovic:
You mean in the `FiberScheduler` constructor? Like this?:

```diff
diff --git a/source/geod24/concurrency.d b/source/geod24/concurrency.d
index 779e5a0..7825a4e 100644
--- a/source/geod24/concurrency.d
+++ b/source/geod24/concurrency.d
@@ -157,6 +157,7 @@ public class FiberScheduler
         this.sem = assumeWontThrow(new Semaphore());
         this.blocked_ex = new FiberBlockedException();
         this.max_parked = max_parked_fibers;
+        thisScheduler(this);
```

Doesn't work for me.

source/geod24/LocalRest.d
```diff
@@ -723,8 +726,9 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
 
         void runNode ()
         {
-            scope node = new Implementation(cargs);
+            is_remote_api = true;
             scheduler = new C.FiberScheduler;
+            scope node = new Implementation(cargs);
```
omerfirmak, line 728: No, I think right here before calling the ctor should work.

omerfirmak: But practically they are the same I guess. Let me give it a try.

AndrejMitrovic: That still seems to segfault for me. Can you give this a shot? :)

Geod24:
I don't see why the change in boolean is warranted ? Switching the order makes sense tho.

omerfirmak:
``` diff
index 8523cf6..643d939 100644
@@ -723,8 +723,9 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
 
         void runNode ()
         {
-            scope node = new Implementation(cargs);
             scheduler = new C.FiberScheduler;
+            C.thisScheduler(scheduler);
+            scope node = new Implementation(cargs);
 
             // Control the node behavior
             Control control;
@@ -2503,3 +2504,39 @@ unittest
     assert(extnode.required() == 42);
     assertThrown!ClientException(extnode.optional());
 }
+
+/// test that runTask works in the constructor
+unittest
+{
+    __gshared bool called;
+    static interface API
+    {
+    @safe:
+        public @property ulong pubkey ();
+    }
+
+    static class MockAPI : API
+    {
+    @safe:
+        this () @trusted
+        {
+            runTask(&callMe);
+        }
+
+        void callMe () @trusted
+        {
+            called = true;
+        }
+
+        public override @property ulong pubkey () { return 42; }
+    }
+
+    scope test = RemoteAPI!API.spawn!MockAPI();
+    scope (exit)
+    {
+        test.ctrl.shutdown();
+        thread_joinAll();
+    }
+    assert(test.pubkey() == 42);
+    assert(called);
+}
```

This works for me. @AndrejMitrovic

AndrejMitrovic:
Yeah it works for me now too. I wonder what I did wrong before, it looked similar to your code. Oh well.

omerfirmak (approved):

PR #117, opened 2021-03-11
Introduce LocalRestConnectionClosed exception

AndrejMitrovic:
This will allow as to at least handle node shutdowns gracefully when using LocalRest in our test-suite. At least it should fix crashes in the tests involving periodic timers.

### DIFF
--- a/source/geod24/LocalRest.d
+++ b/source/geod24/LocalRest.d
@@ -110,6 +110,15 @@ import std.traits : fullyQualifiedName, Parameters, ReturnType;
 import core.thread;
 import core.time;
 
+/// Thrown when the connection to a peer closed (peer node shut down)
+public final class LocalRestConnectionClosed : Exception
+{
+    this () @safe pure nothrow @nogc
+    {
+        super("Connection with peer closed");
+    }
+}
+
 /// Data sent by the caller
 private struct Command
 {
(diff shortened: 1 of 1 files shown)


AndrejMitrovic:
I just realized I can catch it in the Timer routine itself. One second..

AndrejMitrovic:
I made it better now. The `setTimer` will automatically handle the exception if it escapes the delegate. That way we won't have to touch any of the Agora code.

hewison-chris (approved):
LGTM

source/geod24/LocalRest.d
```diff
@@ -2503,3 +2519,69 @@ unittest
...
+    api_2 = node_2.ctrl.listener();
+
+    node_1.test();
+    node_2.ctrl.shutdown();
+    Thread.sleep(1.seconds);  // trigger calling shut down node
+    node_1.ctrl.shutdown();
+
+    size_t count; // up to 3 seconds wait
+    while (!atomicLoad(connection_closed))
+        assert(count < 300); count++; Thread.sleep(10.msecs);
```
omerfirmak, line 2586: So this doesnt need curly braces? D is a weird language.

AndrejMitrovic: You know what happened? The stupid coverage bot failed my PR because 3 lines weren't covered. So I put them all on the same line. And then I removed the braces. I need sleep.

Geod24:
So, is this an issue with client code, or LocalRest ? I'm quite wary of introducing Exceptions that can be handled for specific LocalRest situation.

omerfirmak:
> So, is this an issue with client code, or LocalRest ? I'm quite wary of introducing Exceptions that can be handled for specific LocalRest situation.

IIUC, with this we are almost silently ignoring closed connections. I feel like it should be handled at the client side, but not sure if it is possible to do in agora since we target both LocalRest and vibe.d.

AndrejMitrovic:
> IIUC, with this we are almost silently ignoring closed connections.

We can catch it explicitly and throw a different exception. It will again lead to the node crashing in the tests.

But anyway we need some way of shutting down all nodes at once..

AndrejMitrovic:
Maybe some way of notifying all schedulers to ignore connection failures first, and then issuing the shutdown command..

omerfirmak:
What about a watchdog in the main thread? We can check to see if any node dies prematurely.

PR #119, opened 2021-03-16
Fix blocking in Ctor edge case

omerfirmak:
Since #116, to allow starting tasks in Ctor of nodes,
LocalRest registers the FiberScheduler before starting
the dispatch loop. This might cause the main thread to try
blocking on Fiber level primitives, which is not possible.

isMainThread() should check for a Fiber context rather than
a running FiberScheduler instance and runTask() should be happy
with a scheduler instance even if it is running in the main thread.

### DIFF
--- a/source/geod24/LocalRest.d
+++ b/source/geod24/LocalRest.d
@@ -440,7 +440,7 @@ private bool isMainThread () @trusted nothrow
     // we are in the main thread
     if (scheduler is null)
         scheduler = new C.FiberScheduler;
-    return C.thisScheduler() is null;
+    return Fiber.getThis() is null;
 }
 
 /*******************************************************************************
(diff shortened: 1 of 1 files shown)


source/geod24/LocalRest.d
```diff
@@ -463,7 +463,7 @@ private bool isMainThread () @trusted nothrow
 
 public void runTask (void delegate() dg) nothrow
 {
-    assert(!isMainThread(), "Cannot call this function from the main thread");
+    assert(scheduler !is null, "Cannot call this delegate from the main thread");
```
AndrejMitrovic, line 466: And `sleep` doesn't need this same change right?

omerfirmak: No, sleep should still require a running Fiber

AndrejMitrovic (approved):
Please merge if ready

PR #120, opened 2021-03-16
Try to re-establish closed connections

omerfirmak:


source/geod24/LocalRest.d
```diff
@@ -887,22 +887,41 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
...
-        @trusted nothrow
+        @safe nothrow
     {
-        import std.exception : assumeWontThrow;
-
         this.bind_chn = listener.data;
         this.timeout = timeout;
         assert(bind_chn);
+        if (!this.connect())
+            assert(0, "Failed to connect");
```
Geod24: Maybe only connect on first request ?

omerfirmak: Done

omerfirmak:
Second commit is a workaround to unblock https://github.com/bpfkorea/agora/pull/1737

Geod24:
Regarding second commit, a better solution could be to have a specific version behind which we `assert` / `throw` if the node connect to itself. Something like `Localrest_Pedantic_Checks` if we plan to add more, or just `Localrest_SelfConnect_Check` ? It could be done independently to detect where this is happening (we'll get a stack trace, hopefully).

AndrejMitrovic:
I don't understand the purpose of the first commit. Which problem is it trying to solve? IOW, when do the nodes unintentionally drop LocalRest connections? Is it just a bugfix?

Geod24:
@AndrejMitrovic : There are still a few design issues with LocalRest currently:
- We have two arrays to keep track of incoming / outgoing connections, namely `incoming_conns` and `outgoing_conns`. When we restart those two should be cleaned, but currently the `incoming_conns` are not, which means that the local state changes, but the remote node doesn't get any interruption of service;
- But if we change things to handle disconnections, then we need to add a retry logic to the client. Currently in Vibe.d it just tries to do 2 HTTP requests IIRC.
- Ideally, we should also be able to have a `crash` primitive, or a shutdown without cleanup. This would allow us to test that timeouts are respected. This however would lead to a key difference with Vibe.d: In Vibe, the connection is alive if we received data/a keep-alive within the timeout of the connection, otherwise it needs to be re-established. I think we need something similar in LocalRest on the long run.

Geod24:
Another thing we discussed: Currently one difference in Agora is that `registerListener` is called but doesn't do anything (https://github.com/bpfkorea/agora/blob/f3808b2855a04d6025442f8da0f0711e51e8e39e/source/agora/node/Runner.d#L107-L120) while in LocalRest it is still used (https://github.com/bpfkorea/agora/blob/f3808b2855a04d6025442f8da0f0711e51e8e39e/source/agora/test/Base.d#L1347-L1362).
However it relies on `isOurOwnAddress` which is buggy for non-localrest (with Vibe.d, our own address is e.g. `http://0.0.0.0:2826`, so banning it doesn't help to prevent self-connection).
 The whole idea of `isOurOwnAddress` is flawed, because we could be reached through IPv4, IPv6, DNS, and we might have multiple IP{4,6} addresses anyway.

Geod24:
@omerfirmak : As mentioned on Slack, I don't think we should have the retry logic in the control interface.
The reason for this is that the control interface is a bit special, inasmuch as it exposes a set of functionalities not available to regular clients, and does it solely for the sake of making writing tests easier. So while "regular" code will have to handle disconnects gracefully for Vibe.d or LocalRest, the control interface is typically called at a point where more assumptions are made about the state of the node, and more knowledge exists.

source/geod24/LocalRest.d
```diff
@@ -2281,18 +2288,11 @@ unittest
         {
             assert(this.self is null);
             scope node1 = new RemoteAPI!API(node1Addr);
-            node1.call2();
+            node1.ctrl.shutdown();
```
omerfirmak: @Geod24 , can you check if this is still testing the scenario you intended.

omerfirmak:
Updated, re-try is now configurable and ctrl interface is not affected.
and when compiled with `Localrest_SelfConnect_Check` LocalRest will assert on self connects.

omerfirmak:
Removing the workaround commit since #121 implements a fix.

PR #121, opened 2021-03-22
Fix self conn deadlock

omerfirmak:
When a node goes to sleep and manages to send
a request to itself the request gets queued. Once
it wakes up from the sleep it will try to process
that request synchronous to it's own event loop and lock
itself. Queued messages should be handled async to the
event loop.

### DIFF
--- a/source/geod24/LocalRest.d
+++ b/source/geod24/LocalRest.d
@@ -770,12 +770,14 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
 
                     auto sel_ret = C.select(read_list, write_list, 10.msecs);
 
-                    if (!control.isSleeping())
+                    if (await_msgs.length > 0 && !control.isSleeping())
                     {
-                        foreach (ref await_msg; await_msgs)
-                            handleCommand(await_msg.var.cmd, node, control.filter, await_msg.conn.resp_chn);
+                        scheduler.spawn({
+                            auto prev_await_msgs = await_msgs;
+                            foreach (ref await_msg; prev_await_msgs)
+                                handleCommand(await_msg.var.cmd, node, control.filter, await_msg.conn.resp_chn);
+                        });
                         await_msgs.length = 0;
-                        assumeSafeAppend(await_msgs);
                     }
 
                     if (!sel_ret.success)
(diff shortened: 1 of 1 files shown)


source/geod24/LocalRest.d
```diff
@@ -770,10 +770,13 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
...
-                    if (!control.isSleeping())
+                    if (await_msgs.length > 0 && !control.isSleeping())
                     {
-                        foreach (ref await_msg; await_msgs)
-                            handleCommand(await_msg.var.cmd, node, control.filter, await_msg.conn.resp_chn);
+                        auto await_msgs_duped = await_msgs.dup;
+                        scheduler.spawn({
+                            foreach (ref await_msg; await_msgs_duped)
+                                handleCommand(await_msg.var.cmd, node, control.filter, await_msg.conn.resp_chn);
+                        });
```
Geod24: This defeats the purpose of re-using a buffer. Is there no other way this could be handled ?

source/geod24/LocalRest.d
```diff
@@ -2544,3 +2547,45 @@ unittest
...
+    }
+
+    auto node1 = RemoteAPI!API.spawn!Node();
+    scope (exit)
+    {
+        node1.ctrl.shutdown();
+        thread_joinAll();
+    }
+
+    node1Addr = node1.ctrl.listener();
```
Geod24: Use `cas`

source/geod24/LocalRest.d
```diff
@@ -2544,3 +2547,45 @@ unittest
...
+    __gshared Listener!API node1Addr;
+
+    static class Node : API
+    {
+        private RemoteAPI!API self;
+
+        @trusted:
+        public override void call0 ()
+        {
+            this.self = new RemoteAPI!API(node1Addr);
```
Geod24: Use `atomicLoad`

source/geod24/LocalRest.d
omerfirmak: I dont think there is, without some overkill data structures. But I think we are better off reallocating after each sleep rather than duplicating and reusing the buffer.

Geod24: Makes sense. Let's go with this.

omerfirmak:
All done

source/geod24/LocalRest.d
```diff
@@ -770,12 +770,14 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
 
                     auto sel_ret = C.select(read_list, write_list, 10.msecs);
 
-                    if (!control.isSleeping())
+                    if (await_msgs.length > 0 && !control.isSleeping())
                     {
-                        foreach (ref await_msg; await_msgs)
-                            handleCommand(await_msg.var.cmd, node, control.filter, await_msg.conn.resp_chn);
+                        scheduler.spawn({
+                            auto prev_await_msgs = await_msgs;
```
AndrejMitrovic, line 776: In the future I think it's enough to just slice the array. Same effect but gets rid of the temporary variable.

AndrejMitrovic (approved):

PR #127, opened 2021-08-11
Implement critical sections

### LINKED ISSUE #126: Implement a `suspend()` primitive to complement `sleep()`
Currently, we have the ability to make a node `sleep` for a certain time.
However, while a node is `sleep`ing, its timers and scheduled tasks are still being processed, which can come as a surprise for users.
We should have an extra primitive to allow suspending a node. Naming wise, I'm not sure which name fits which scenario best, so `sleep` could be made to actually make the node unresponsive (except for other commands) and `suspend` added, or other names completely could be used.

omerfirmak:
This will help implementing complete suspension of
LocalRest nodes (incl Timers, Tasks).

### DIFF
--- a/source/geod24/LocalRest.d
+++ b/source/geod24/LocalRest.d
@@ -130,6 +130,8 @@ private struct TimeCommand
     Duration dur;
     /// Whether or not affected messages should be dropped
     bool drop = false;
+    /// Whether or not all tasks should be suspended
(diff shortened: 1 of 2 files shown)


omerfirmak:
Closes #126

Geod24 (approved):
Neat

PR #129, opened 2021-09-14
Allow casting between RemoteAPI templates

omerfirmak:
where downcasting the API is possible

### DIFF
--- a/source/geod24/LocalRest.d
+++ b/source/geod24/LocalRest.d
@@ -921,8 +921,9 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
 
     ***************************************************************************/
 
-    public this (Listener!API listener, Duration timeout = 5.seconds)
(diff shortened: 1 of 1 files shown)


source/geod24/LocalRest.d
```diff
@@ -921,9 +921,10 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
 
     ***************************************************************************/
 
-    public this (Listener!API listener, Duration timeout = 5.seconds)
+    public this (IncomingAPI) (Listener!IncomingAPI listener, Duration timeout = 5.seconds)
         @trusted nothrow
     {
+        static assert(is(IncomingAPI : API));
```
Geod24: Error message ?

omerfirmak: Updated, how about that?

Geod24: Works too, although static assert provides a better UX.

PR #130, opened 2021-12-07
Extend timer interface

mkykadir:
Include more (useful) interfaces from Vibe.d's timer

- `createTimer`: Creating timer without arming
- `pending`: State of the timer after arming (if waiting to be fired)

### DIFF
--- a/source/geod24/LocalRest.d
+++ b/source/geod24/LocalRest.d
@@ -502,7 +502,7 @@ public void sleep (Duration timeout) nothrow
 
 *******************************************************************************/
 
-public Timer setTimer (Duration timeout, void delegate() dg,
+public Timer setTimer (Duration timeout, void delegate() @safe nothrow dg,
     bool periodic = false) nothrow
 {
     assert(scheduler !is null, "Cannot call this delegate from the main thread");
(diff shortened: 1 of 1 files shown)


source/geod24/LocalRest.d
```diff
@@ -567,6 +587,12 @@ public final class Timer
             scheduler.schedule(&run);
         }
     }
+
+    /// True if timer is yet to fire
+    @property bool pending () @safe nothrow
+    {
+        return this._pending;
```
mkykadir, line 605: Maybe `&&` with `running` or setting `_pending` to `false` is required at L558

Geod24:
I don't understand why we need `createTimer` ? The call site will have to select which function to call anyway, so it's not an abstraction.

mkykadir:
Just for convenience actually. One might want to create a timer in ctor and later manage it. `setTimer` creates and arms a timer which might be limiting for some cases. (Reentrant functions)

source/geod24/LocalRest.d
hewison-chris: I would go with the second option so the state of `this.pending` is updated as it could be accessed not using this property and have the wrong state.

hewison-chris (changes requested):
LGTM, small addition as you suggested.

omerfirmak (approved):

hewison-chris (commented):
LGTM

PR #134, opened 2022-02-02
Timer: Allow timer rearm in own handler

### LINKED ISSUE #133: Timer `rearm` behavior is different from Vibe.d's behavior
Vibe.d's `rearm` allows to set timer within the handler, thus following is valid and works every 2 seconds;

```
Timer timer;
void printTime()
@safe nothrow {
  import std.datetime;
  logInfo("The time is: %s", Clock.currTime());
  timer.rearm(2.seconds);
}

...
timer = createTimer(toDelegate(&printTime));
timer.rearm(2.seconds);
```

LocalRest timer doesn't arm the timer within its handler due to running check condition;
https://github.com/Geod24/localrest/blob/7fc2386fe75e5b944090d9851f025c1882649204/source/geod24/LocalRest.d#L592-L595

mkykadir:
Vibe.d timers allow rearming within the timer handler itself, this makes
both timer times behave more similar.

Fixes #133

### DIFF
--- a/source/geod24/LocalRest.d
+++ b/source/geod24/LocalRest.d
@@ -545,12 +545,15 @@ public final class Timer
     private bool running;
     // Whether this timer is waiting for timeout
     private bool _pending;
+    // Whether this timer is rearmed in its handler
+    private bool rearmed;
 
     @safe nothrow:
     public this (void delegate() @safe nothrow dg)
     {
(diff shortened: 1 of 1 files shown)


mkykadir:
Moved here -> https://github.com/bosagora/localrest/pull/1

